PR #1, opened 2020-12-12
Programmable block types and mapping on different target architectures

mariobaldi:
I have changed the Programmable blocks section with the type declarations for the v4 architecture (I have kept the previous declarations as a subsection).

Moreover, I have added an appendix explaining how a compiler can map a program written according to the v4 architecture on a target that looks like v3 or a target that has a single programmable block with single parser, control and deparser.

The markup text has not been compiled, so there may be syntax errors.


jnfoster:
Mario, can you sign the ONF CLA or make sure one is on file for Pensando with your GitHub ID included?

PR #5, opened 2021-03-15
Add proposed mirror_packet extern function and example program using it

jfingerh:


jafingerhut:
Merging for convenience of reference, NOT because it is considered final for release of PNA specification.  Still work to do to decide whether this or some modified version becomes part of released PNA specification.

PR #21, opened 2021-05-16
Modify architecture diagrams

jnfoster:
Shade "Message Procession" as "Planned Extension" rather than "P4 Programmable"


jafingerhut (approved):
I would say "looks good to me", but I am assuming at the moment it will based on our chat :-)

PR #44, opened 2022-06-01
Function to set expire timer of an added entry.

roop-nvda:
Per discussion in the call last week, I have uploaded this PR with the proposed function to set expiry timer along with some examples.

### DIFF
--- a/pna.p4
+++ b/pna.p4
@@ -744,31 +744,27 @@ extern void update_expire_info(
 
 
 // Set the expire time of the matched entry in the table to the value
-// specified in the parameter expire_time_profile_id, if the value
-// parameter is equal to the data parameter. If the value and data
-// parameters are not equal, the function has no effect.
+// specified in the parameter expire_time_profile_id, if condition
+// in the first parameter evaluates to true. Otherwise, the function
+// has no effect.
 //
-// @param D          Tuple of fields of data to match with value.
-// @param data       Data that is matched to the value parameter
-//                   to determine if the expire time will be set.
-// @param value      Value that is matched with corresponding
-//                   data parameter.
-// @expire_time_profile_id
+// @param condition  The boolean expression to evaluate to determine
+//                   if the expire time will be set.
+// @param expire_time_profile_id
 //                   The expire time to set for the matched entry,
 //                   if the data and value parameters are equal.
 //
 // Examples:
-// set_expire_time_if({hdr.tcp.flags, meta.direction},
-//                    {TCP_FLG_SYN, OUTBOUND},
+// set_expire_time_if(hdr.tcp.flags == TCP_FLG_SYN &&
+//                    meta.direction == OUTBOUND,
 //                    tcp_connection_start_time_profile_id);
-// set_expire_time_if(hdr.tcp.flags, TCP_FLG_ACK,
+// set_expire_time_if(hdr.tcp.flags == TCP_FLG_ACK,
 //                    tcp_connection_continuation_time_protile_id);
-// set_expire_time_if(hdr.tcp.flags, TCP_FLG_FIN,
+// set_expire_time_if(hdr.tcp.flags == TCP_FLG_FIN,
 //                    tcp_connection_close_time_profile_id);
 
-extern void set_entry_expire_time_if<D>(
-    in D data,
-    in D value,
+extern void set_entry_expire_time_if(
+    in bool condition,
     in ExpireTimeProfileId_t expire_time_profile_id);
 
 
--- a/pna.p4
+++ b/pna.p4
@@ -743,6 +743,35 @@ extern void update_expire_info(
     in ExpireTimeProfileId_t expire_time_profile_id);
 
 
+// Set the expire time of the matched entry in the table to the value
+// specified in the parameter expire_time_profile_id, if the value
+// parameter is equal to the data parameter. If the value and data
+// parameters are not equal, the function has no effect.
+//
+// @param D          Tuple of fields of data to match with value.
+// @param data       Data that is matched to the value parameter
+//                   to determine if the expire time will be set.
+// @param value      Value that is matched with corresponding
+//                   data parameter.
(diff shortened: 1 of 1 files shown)


pna.p4
```diff
@@ -743,6 +743,35 @@ extern void update_expire_info(
...
+// @param value      Value that is matched with corresponding
+//                   data parameter.
+// @expire_time_profile_id
+//                   The expire time to set for the matched entry,
+//                   if the data and value parameters are equal.
+//
+// Examples:
+// set_expire_time_if({hdr.tcp.flags, meta.direction},
+//                    {TCP_FLG_SYN, OUTBOUND},
+//                    tcp_connection_start_time_profile_id);
```
jfingerh, line 763: If the new parameters for `set_expire_time_if` were a single additional parameter `in bool condition`, then the example above could be written as shown below:
```
set_expire_time_if((hdr.tcp.flags == TCP_FLG_SYN) && (meta.direction == OUTBOUND),
                              tcp_connection_start_time_profile_id);
```

jfingerh:
@roop-nvda Similar to the discussion we had on `add_entry_if` in this issue https://github.com/p4lang/pna/pull/43 my main comment would be to express my strong preference that if we want to create extern functions that have explicit parameters to make the side effects conditional, that we do so with a single `in bool condition` parameter rather than two parameters `val1` and `val2` that restrict the condition to `val1 == val2`.

I have added another comment to your PR showing how one of your example calls to `set_expire_time_if` examples could be rewritten if it has such a `bool` first parameter.

As for the discussion on https://github.com/p4lang/pna/pull/43, it seems to me that doing it this way lets one express the functionality you want, as well as more general conditional expressions, e.g. `(a < 5)`.  If a target device cannot support `(a < 5)`, then that is up to the compiler for that target to say it is too complex, preferably with links to documentation on what kinds of expressions that target supports in such conditions.

jafingerhut:
So suppose in the kind of target that you are thinking of, the back end of the P4 compiler checked the conditional expression's IR to see if it met the following conditions.  This seems like a fairly straightforward thing for a back-end to check in time linear in the number of IR nodes of the expression, i.e. it would be very fast and simple:

+ verify that all leaves are metadata variables or constants
+ verify that all parent nodes in the IR of the leaves are `==` operators
+ verify that all other nodes in the IR tree of the expression are `&&` operators (`&&` is associative, so the shape of the IR nodes in the tree does not matter)

If those were true, then it is a straightforward linear time operation (at compile time, not run time in the data plane) to know that any expression `(a1 == b1) && (a2 == b2) && ...` is equivalent to `{a1, a2, ...} == {b1, b2, ...}`.

If those conditions were not true, the back end of the target you have in mind would reject the condition as unsupported (for that target device).

Compilers for particular P4 targets are allowed to reject programs as too complex.  They do it for any number of reasons all the time, differently for each back end target device, depending on the details of the target's capabilities.  In a case like this, the error message for rejected expressions could even be very helpful in the back end: "supported conditional expressions must be of the form `(a1 == b1) && (a2 == b2) && ...` where all a's and b's are either header or metadata fields, or constants.

The restrictions of such a back end need not restrict OTHER P4 compiler back ends for other targets that might have different restrictions, though.  Is that 100% portable?  No, of course it isn't.  But we know that different targets will have restrictions that differ between them like: maximum number of recommended or hardware-permitted recirculate operations per packet, maximum number of table apply operations supported in a single pass, etc.  I don't think anyone involved in PNA wants to try to mandate particular numerical values for those things that all targets must comply with -- they will vary today, and they will likely vary across different devices from a single vendor, too.

roop-nvda:
Apologize if it was not clear, but I was not suggesting that numerical values of the size of tuple argument of `set_expire_time_if` be mandated in PNA. I merely gave an example with a fictitious target -- Acme -- to illustrate that the error message was simple for the programmer. Different targets are indeed expected to support different sized tuples for matching, but the semantic of a match is simple -- `(data[0] == value[0]) && (data[1] == value[1]) ... && (data[N-1] == value[N-1])`-- and obvious across the class of targets.

When compiling for bmv2 simple switch backend with a conditional in a match action we get an error message implying that only C/C++ ternary assignment type conditionals are allowed -- `only conditionals like if (a > 5) { a = 5;} are allowed in match actions for this target`.  Different targets expressing this and other classes of conditionals that are supported/unsupported in an error message in a way that helps the programmer fix the program can get cumbersome.

jfingerh:
Maybe I am answering a different concern than you are worried about, but note that the bmv2 back end today can support assignment statements like this inside of P4 actions, as demonstrated by the attached program:
```
    action a1() {
        b1 = (hdr.ethernet.srcAddr == hdr.ethernet.dstAddr) && (hdr.ethernet.etherType == 0x0800);
    }
```
since it can implement that with no changes, I suspect that it should also be able to implement an action like this with the only change required to be defining and implementing an extern `set_expire_time_if` that has a first parameter with type `in bool`:
```
set_expire_time_if((hdr.tcp.flags == TCP_FLG_SYN) && (meta.direction == OUTBOUND),
                   tcp_connection_start_time_profile_id);
```
[v1model_bool_expressions_supported1.p4.txt](https://github.com/p4lang/pna/files/8995742/v1model_bool_expressions_supported1.p4.txt)

jafingerhut:
@roop-nvda Here is another angle on the question.  Suppose we did say we were only going to implement your proposed version of `set_expire_time_if`, where the first two parameters must be equal in order for the side effect to be performed.

Then it is perfectly legal P4 syntax to write this:
```
    set_expire_time_if(a < 5, true, tcp_connection_start_time_profile_id);
```
because `a < 5` is an expression that evaluates to a type `bool` value, and that is the same type as the second parameter.

So trying to require two parameters that must be equal, trying to restrict the generality of conditional expressions that can be expressed, actually does not restrict it at all.

roop-nvda:
@jfingerh  -- Agreed. I will amend and resend the PR.

PR #46, opened 2022-06-27
Remove send_to_vport() from remaining example programs

jfingerh:
send_to_vport() extern function was removed a while ago, but was not
removed from all example programs at that time.  Catching up with that
months-old change in the examples with this commit.


jfingerh:
Merging this in quickly, to base further proposed changes upon.  As mentioned in the commit message, this is really just catching some of the examples up to a change made quite a while ago in the pna.p4 #include file.

PR #48, opened 2022-06-27
Proposed changes to pna.p4 and one example PNA P4 program

jfingerh:
The others would need to be changed similarly if this proposal goes forward.

### DIFF
--- a/examples/pna-demo-L2-two-tables.p4
+++ b/examples/pna-demo-L2-two-tables.p4
@@ -77,7 +77,7 @@ control MainControlImpl(
         drop_packet();
     }
     action L2_send_to_port (PortId_t port_id) {
-        send_to_port(port_id);
+        ostd.egress_port = port_id;
     }
 
     // In this demo program, I have chosen to use the same action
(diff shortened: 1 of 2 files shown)


jfingerh:
The only change here is to add an intrinsic metadata field in place of the current `send_to_port(p)` extern function.

If people prefer the current extern function style for sending packets to ports, dropping packets, etc. then this PR can be closed.

shirshyad:
@jfingerh could we pls add git-issue if there is one and point to that in the git-description or description as to why / motivation for this proposal.

jfingerh:
The motivation for this proposal was a discussion topic that Dan Talayco talked about at a PNA meeting perhaps a month or two ago.

I am fine if we adopt an intrinsic metadata style for such things, and I am fine if we adopt a 'use extern function to set these values, without corresponding get functions'.

They both achieve similar results, but not quite identical.  Part of the PNA group's work is to pick something here, and run with it.

shirshyad:
> The motivation for this proposal was a discussion topic that Dan Talayco talked about at a PNA meeting perhaps a month or two ago.
> 
> I am fine if we adopt an intrinsic metadata style for such things, and I am fine if we adopt a 'use extern function to set these values, without corresponding get functions'.
> 
> They both achieve similar results, but not quite identical. Part of the PNA group's work is to pick something here, and run with it.

Would it not be possible to keep send_to_port extern function and where intrinsic meta is to be used such as in table key, ```ostd.egress_port``` can be specified.  If there is a reason to remove extern function and how it helps in git-description, then it will help. I have missed Dan T 's talk about removing the extern function.

PR #49, opened 2022-07-15
Adding packet length to Meter execute and Counter count methods for targets that may need it

usha1830:
Proposing to add an optional parameter to specify the packet length for Meter extern's execute method and and Counter extern's count method.

If we can add this parameter as optional to the pna specification, it would help DPDK target to comply with the spec. For targets that don’t need packet length, it can be simply ignored.

### DIFF
--- a/pna.p4
+++ b/pna.p4
@@ -444,7 +444,7 @@ enum PNA_CounterType_t {
 
 extern Counter<W, S> {
   Counter(bit<32> n_counters, PNA_CounterType_t type);
-  void count(in S index);
+  void count(in S index, @optional in bit<32> increment);
 }
 // END:Counter_extern
 
@@ -475,12 +475,12 @@ extern Meter<S> {
   // Use this method call to perform a color aware meter update (see
   // RFC 2698). The color of the packet before the method call was
   // made is specified by the color parameter.
-  PNA_MeterColor_t execute(in S index, in PNA_MeterColor_t color);
+  PNA_MeterColor_t execute(in S index, in PNA_MeterColor_t color, @optional in bit<32> pkt_len);
 
   // Use this method call to perform a color blind meter update (see
   // RFC 2698).  It may be implemented via a call to execute(index,
   // MeterColor_t.GREEN), which has the same behavior.
-  PNA_MeterColor_t execute(in S index);
+  PNA_MeterColor_t execute(in S index, @optional in bit<32> pkt_len);
 }
 // END:Meter_extern
 


jfingerh:
@usha1830 Do you still want this PR?  I think we discussed it privately several weeks ago, but I haven't followed up with you since then.

pna.p4
```diff
@@ -444,7 +444,7 @@ enum PNA_CounterType_t {
 
 extern Counter<W, S> {
   Counter(bit<32> n_counters, PNA_CounterType_t type);
-  void count(in S index);
+  void count(in S index, @optional in bit<32> increment);
```
shirshyad, line 447: Optional increment is not applicable to packet counter. It may apply to byte counter.
Can we revise this and look into a way address application of optional increment only on byte counters ?

shirshyad, line 447: Also should we look into adding a new method instead of modifying existing one ?

shirshyad (changes requested):

pna.p4
apinski-cavium, line 447: Is there a reason why you use @optional here when doing the following should also work:
```
void count(in S index);
void count(in S index, in bit<32> increment);
```
I don't see why @optional is needed when you can do it similarly via overloading.

apinski-cavium, line 447: Is there a reason why you use @optional here when doing the following should also work:
```
void count(in S index);
void count(in S index, in bit<32> increment);
```
I don't see why @optional is needed when you can do it similarly via overloading.

usha1830: We will handle this in p4c and expect the packet length only in case of byte counters. We can choose to either reject the programs with packet length parameter passed to packet counters or ignore that parameter with a warning.

usha1830: Yes.

usha1830: Yes. We will remove @optional.

usha1830:
> @usha1830 Do you still want this PR? I think we discussed it privately several weeks ago, but I haven't followed up with you since then.

@jfingerh I will close this PR. 
As per our discussion earlier, we will create new methods which take packet length and will also remove the @optional.

In case of counter 'count' method, an overloaded function would work as @apinski-cavium pointed out. However, in case of Meter 'execute' , we may have to change the method name to avoid the conflict between the below two signatures. p4c doesn't seem to resolve overloaded methods based on type.

` PNA_MeterColor_t execute(in S index, in PNA_MeterColor_t color); `
conflicts with 
`PNA_MeterColor_t execute(in S index,  in bit<32> pkt_len);`

pna.p4
shirshyad: For P4c or backend to reject optional increment argument, instance of Counter type need to be qualified so that P4C/ backend knows it a byte counter instance versus packet counter instance. Is this something that's possible with the way the Counter<W,S> class is defined ? If not, I am not sure how we can error out increment argument if the method is invoked on packet counter instance. (not on byte counter object or on packet_and_byte counter object)

jfingerh: Whether a counter instance is packet, byte, or packet & byte, is known during compile time based upon the arguments to the constructor call that created the instance.  The P4 compiler should have all of the information required to know which kind each counter instance is.

usha1830:
Closing this PR as discussed.
@kamleshbhalui is working on a new PR for adding new methods for count and execute for dpdk.

PR #51, opened 2022-08-15
Add two variants of PNA

jfingerh:


shirshyad:
High Level Comments

- To me based on IPSEC requirement which requires packet parsing once before decrypt action and once again post decrypt action,  two architectures are being proposed. I am not sure this approach of architecture variants based on offload functionality will set a precedence that can lead to more architectures in the future either because offloads are implemented  uniquely by various vendors or due to constraints of tooling. If a  common abstraction is not possible to arrive at, then should we leave IPSEC offload out of standardization effort and let vendor specific extern/s express IPSEC at the cost of portability ?
- It would help to hear from the end user of P4 who would use NIC from multiple vendors.  We should consider their inputs before we checkin two version of PNA architectures.
- If there is conclusion that more than one architectures is unavoidable, then should we just describe abstract model that merges both PNA1 and PNA2 and state in the specification that every vendor need not implement all the blocks of the abstract model and  hence call out that portability is impacted.

pna1.p4
```diff
@@ -0,0 +1,54 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#ifndef __PNA1_P4__
+#define __PNA1_P4__
+
```
shirshyad, line 18: Should we not edit pna.p4 instead of adding pna1.p4 ? We may  remove PreControl from pan.p4 and a new pna2.p4 file that expresses second architecture with pre-preparer and pre-control.

jafingerhut:
I plan to close this PR as soon as I determine that there is nothing I want to salvage from it.  It uses an approach that I have given up pursuing for inclusion in PNA.

jfingerh:
I have looked through the changes in this PR, and did not find any changes that were unrelated to the 2-parser/2-control vs. 1-parser/1-control idea, so closing this PR as abandoned.

Issue #45, opened 2022-06-27
Proposed new externs ExactMap and TernaryMap


jfingerh:
This was first discussed in a public PNA meeting on 2022-Jun-13.

The original motivation for the new proposed externs is to have constructs that are in some ways similar to P4 tables, but:

(a) their `lookup` method can be called from inside the body of a P4 action.  The P4_16 language spec explicitly disallows `table_name.apply()` calls within P4 action calls today, and always has since version 1.0.

(b) they simply "return a value" rather than execute an arbitrary P4 action.

Attached is a P4_16 source file pna_extern_defns.p4.txt that defines the extern objects.  Something like this would be added to the pna.p4 #include file if this proposal goes forward.

Below is an edited version of a file that I used to explain the new extern's behavior in the first meeting discussing it, plus some notes of discussion we had during that meeting.

The originally proposed name of ExactMap was ExactMatchLookupTable.
Not only does this seem too long, but putting "table" anywhere in its
name makes it more confusing to talk about, since it is _not_ a P4
table (despite the similarities).  ExactMap was proposed as a shorter
name that avoids the "table" word confusion.

Here are some examples of how to instantiate and call the originally
proposed extern object originally proposed, as defined in the file
pna_extern_defns.p4:

```
    // Example of ExactMap with type bit<W> for key type and return
    // value type.

    // instantiation of an instance named `emap1`, in a control before
    // the apply { ... } block

    ExactMap<bit<8>, bit<16>, _>(
            size = 128,
            default_value = 42)  // default value returned for all other keys
        emap1;

    bit<16> v;

    apply {
        // ... other code here ...
        v = emap1.lookup({hdr.ipv4.protocol);

        // A target _might_ support multiple lookup() operations on
        // the same instance per packet, or it might limit the number
        // of lookup() operations to 1 per packet, 2 per packet, etc.
        // That is a target-specific decision.  The most portable P4
        // code would limit itself to 1 lookup operation per instance,
        // per packet, just as P4 developers do today for P4 table
        // apply() calls.
        
        //w = emap1.lookup(hdr.ipv4.srcAddr[7:0]);

        // ... more other code here ...
    }
```

For the purposes of understanding the behavior of the code above, it
should be the same as the code below, which does not use the new
extern.
    
```
    // Code that behaves the same as the above, and has a very similar
    // control plane API

    bit<16> v;

    action emap1_set_value (bit<16> val) { v = val; }

    table emap1_table {
[...]

PR #52, opened 2022-08-22
Add ExactMap and TernaryMap extern definitions to pna.p4

jfingerh:


jfingerh:
See https://github.com/p4lang/pna/issues/45 for some notes on an earlier discussion of these externs.

mbudiu-vmw:
I think this should be a separate library file, which can be included in pna.p4 if it is deemed to be standard.

mbudiu-vmw:
This was discussed in the LDWG in May 2021: https://github.com/p4lang/p4c/pull/2739

jfingerh:
"I think this should be a separate library file, which can be included in pna.p4 if it is deemed to be standard."  Making it part of the PNA standard is exactly the reason this is being proposed and discussed.  It might not meet with wide enough approval to become part of pna.p4, but that is my hope.

Of course we could make N optional parts of PNA.  If that is met with reasonably wide agreement, we might do that.  In practice, the larger N is, and the more often P4 developers use the optional features that are not implemented by all target devices, the less portable their programs will be.  That doesn't automatically make every optional thing "bad", just less desirable than something all vendors agree to implement.

jfingerh:
@apinski-cavium Is it quick for you to point out the way in which this code violates the current P4 language spec?

@mbudiu-vmw If you or Andrew have any thoughts on changes to this proposal that would enable an argument like `const_entries` to be specified, I am curious to learn about them.  I would prefer one that did NOT require defining a separate constructor for each different number of entries, as that seems doomed to fail whenever someone wants to specify one more entry than the maximum someone picked when defining the constructors.

As you can tell from this proposal, I am OK if a proposal does not enable the p4c front/mid-end code to type check the value of const_entries, and any such type checking is relegated to the back end code of the target.  Once that checking is done for one target in open source, other implementations could hopefully reuse it.

apinski-cavium:
So it looks like I was wrong about _ types. See 7.2.13 Don't care types but I am still not 100% understand the semantics here.
The example given in Section 17.1 only allows them on generics type while doing type unification on the generic type's paramaters (basically an anymous generic paramater type).  When this was added, there was a good example and what seemed like a better explaination of it but it was changed and the example was moved and explanation changed.

jafingerhut:
@apinski-cavium Do you mean that Section 7.2.13, and/or Section 17.1, were different in earlier versions of the spec that had a different explanation/example of _ types?  I did a quick spot-check of the corresponding sections back to version 1.0.0 of the spec (available in the "Archives" section of this page https://p4.org/specs/) but did not notice any differences there.

apinski-cavium:
> @apinski-cavium Do you mean that Section 7.2.13, and/or Section 17.1, were different in earlier versions of the spec that had a different explanation/example of _ types? I did a quick spot-check of the corresponding sections back to version 1.0.0 of the spec (available in the "Archives" section of this page https://p4.org/specs/) but did not notice any differences there.

I looked into the git history (rather than published version of the spec) on this one to find where it changed. It changed when https://github.com/p4lang/p4-spec/commit/84db45872a112fd267485a2b16db51174b7f09af commit was pulled in. The example was never right to begin with anyways (because control type definition vs control definition). 
Note I still think there needs to be a clarification in the language dealing with extern constructors and externs functions because I think the current definition only allows it for (package, control and parser and action) type definitions rather than function calls.

mbudiu-vmw:
The right solution is to extend the p4 type system to support generic lists. The entries is a list literal. The braces notation is somewhat overloaded now: it is used for tuples and structs too.

jafingerhut:
@apinski-cavium Do you think that with a clarification in the language spec, that this proposed extern definition would be more clearly according to that clarified definition?

mbudiu-vmw:
Why don't we actually add List or Vector types and literals to the language? May not be that difficult. We can talk about it at the next design working group. I will try to prototype something to make sure it works.

jafingerhut:
@mbudiu-vmw @jnfoster How likely do you think extending the P4 type system to support generic lists would be in, say 3 to 6 months?  I know that of course the answer depends upon how much effort is put into it, but wondering if you have some kind of guesstimate.  If a smaller change to the language spec would make this PR legal using `_` as it does (or some small variation), I'm content defining it that way for now.

mbudiu-vmw:
This could be done in a couple of days of work if there are no unexpected surprises.

jnfoster:
Agree. Just adding polymorphic lists would be an easy task.

mbudiu-vmw:
Here is a possible example: https://github.com/p4lang/p4c/pull/3520.
Check out the test files.

mbudiu-vmw:
p4c actually accepts the vectors of tuples, but there's a bug in the tuple elimination pass in the midend which replaces tuples with structs, that one causes the errors reported.

mbudiu-vmw:
I have amended the PR to accept the vectors of tuples correctly.

jafingerhut:
Commit 2 on this PR updates the proposed new externs so that they use the new `list` type that may soon be added to the P4_16 language.  For more information, see this PR for the language spec: https://github.com/p4lang/p4-spec/pull/1168, which contains a link to the PR for the p4lang/p4c repo that adds this feature.

As of 2022-Nov-12, the PR for p4lang/p4c might give a CompilerBug when trying to compiler this code, but Mihai Budiu, the author of that PR, has been notified of this, and I expect he may be able to update that PR to accept this code.  It seems best to only merge in this PR after the open source p4c compiler accepts it.

examples/pna-example-exactmap-ternarymap.p4
```diff
@@ -0,0 +1,248 @@
...
+             {0x00ffff, 0x2},
+             2}
+        },
+        default_value = 42)  // default value returned for all other keys
+    tmap1;
+
+    TernaryMap<tmap2_key, tmap2_val>(
+        size = 1024,
+        largest_priority_wins = true,
+        initial_entries =
```
jfingerh, line 170: This `initial_entries` parameter is intended to correspond with the new ability to define initial entries (without ` const` qualifier before the `entries` table property name) proposed in this PR for the language spec: https://github.com/p4lang/p4-spec/pull/1180

Here I do not see an easy way to make the per-entry `const` property or the per-entry `priority` optional, so they are required for each entry.

pna.p4
```diff
@@ -545,6 +545,224 @@ extern Digest<T> {
...
+}
+
+struct exactmap_initial_entry_t<K,V> {
+    bool const_entry;
+    K key;
+    V val;
+}
+
+/**
+ * The type K must be a struct type.  Each of its members becomes a
```
psivanup, line 561: K as struct type: It is possible for user to pass any type for this template parameter. There is no restrictions on number of fields, nesting the struct type. Hope, it is up to the target to restrict on these semantics?

examples/pna-example-exactmap-ternarymap.p4
```diff
@@ -0,0 +1,319 @@
...
+            // TernaryMap.
+            my_field: exact;
+        }
+        actions = { lookup; }
+        default_action = lookup(42);
+    }
+*/
+    ExactMap<emap1_key, bit<16>>(
+        size = 1024,
+        const_entries = (list<exactmap_const_entry_t<emap1_key, bit<16>>>) {
```
psivanup, line 121: Isn't this initialization expected to work without cast ((list<exactmap_const_entry_t<emap1_key, bit<16>>>))?

psivanup (commented):
Looks fine.

pna.p4
jfingerh: Yes, targets can limit the total size and/or type of fields within the struct type K, e.g. if the user wants a 10,000 bit key, that is too much for most targets.

examples/pna-example-exactmap-ternarymap.p4
jfingerh: The current p4c compiler implementation of this feature might NOT work without the `(list<...>)` there.  Maybe it will be enhanced later to work without that there, but it isn't there yet.

pna.p4
```diff
@@ -545,6 +545,224 @@ extern Digest<T> {
...
+    EntryPriority_t priority;
+    K key_val;
+    K key_mask;
+    V val;
+}
+
+
+/**
+ * The type K must be a struct type.  Each of its members becomes a
+ * separate key field of the ExactMap instance, with the member name
```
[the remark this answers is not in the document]
jfingerh: Good catch!  I will fix this in a later commit soon.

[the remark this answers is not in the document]
jfingerh: Fixed.

thomascalvert-xlnx (commented):
These are so similar to existing tables (the comments even show the equivalence), I think it would be nicer if we could somehow implement this using a table property. A couple ideas:
```
    table emap2 {
        key = {
            val1: exact;
            val2: exact;
        }
        // notice lack of "actions =" line here
        // lookup would be a special action recognised by the compiler
        default_action = lookup(emap2_val);
        // alternatively, we could introduce this new table property:
        direct_value = emap2_val;
    }
    (...)
    bit<4>  my_val1 = ...;
    bit<12> my_val2 = ...;
    emap2_val my_resp = emap2.lookup(my_val1, my_val2);
```
Personally I have a slight preference for `direct_value`, as this is an action-less table (and so `default_action` doesn't really make sense).

The underlying BCAM/TCAM primitives aren't something which is specific to a NIC as far as I know.

----------

>> I think this should be a separate library file, which can be included in pna.p4 if it is deemed to be standard.

> Of course we could make N optional parts of PNA. If that is met with reasonably wide agreement, we might do that. In practice, the larger N is, and the more often P4 developers use the optional features that are not implemented by all target devices, the less portable their programs will be. That doesn't automatically make every optional thing "bad", just less desirable than something all vendors agree to implement.

The flip side is that once something is in `pna.p4` it becomes very hard to change our minds down the road. However going the other way is easier, i.e. moving an optional extern into the main `pna.p4` file.

jafingerhut:
> These are so similar to existing tables (the comments even show the equivalence), I think it would be nicer if we could somehow implement this using a table property. A couple ideas:
> 
> ```
>     table emap2 {
>         key = {
>             val1: exact;
>             val2: exact;
>         }
>         // notice lack of "actions =" line here
>         // lookup would be a special action recognised by the compiler
>         default_action = lookup(emap2_val);
>         // alternatively, we could introduce this new table property:
>         direct_value = emap2_val;
>     }
>     (...)
>     bit<4>  my_val1 = ...;
>     bit<12> my_val2 = ...;
>     emap2_val my_resp = emap2.lookup(my_val1, my_val2);
> ```
> 
> Personally I have a slight preference for `direct_value`, as this is an action-less table (and so `default_action` doesn't really make sense).
> 
> The underlying BCAM/TCAM primitives aren't something which is specific to a NIC as far as I know.
> 
> > > I think this should be a separate library file, which can be included in pna.p4 if it is deemed to be standard.
> 
> > Of course we could make N optional parts of PNA. If that is met with reasonably wide agreement, we might do that. In practice, the larger N is, and the more often P4 developers use the optional features that are not implemented by all target devices, the less portable their programs will be. That doesn't automatically make every optional thing "bad", just less desirable than something all vendors agree to implement.
> 
> The flip side is that once something is in `pna.p4` it becomes very hard to change our minds down the road. However going the other way is easier, i.e. moving an optional extern into the main `pna.p4` file.


I am perfectly OK if we end up just creating more general P4 tables for this purpose, but note one of the original motivations for this in #45, repeated here:

(a) their lookup method can be called from inside the body of a P4 action. The P4_16 language spec explicitly disallows table_name.apply() calls within P4 action calls today, and always has since version 1.0.

I believe it was Roop who asked recently about this same issue, and I pointed out to him this limitation in the current P4 language spec, and our use case for wanting to call these externs within actions, and I also pointed out the following issue on the public p4-spec Github repo where the P4 language spec is discussed: https://github.com/p4lang/p4-spec/issues/947

Roop said he is motivated to raise this issue again, hopefully in the next P4 language design work group meeting on 2022-Dec-05, and I urge you to also attend that meeting to make your case, or support Roop however you can, if you want such a change made in the spec.

examples/pna-example-exactmap-ternarymap.p4
```diff
@@ -0,0 +1,319 @@
...
+        default_action = drop;
+    }
+    apply {
+        // Note that table L2_fwd is one table accessed by packets
+        // being processed in both the NET_TO_HOST and the HOST_TO_NET
+        // directions.
+
+        // As far as the control plane API is concerned, there is
+        // exactly one table.
+        L2_fwd.apply();
```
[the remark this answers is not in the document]
jfingerh: Agreed.  I wanted to get a variety of instantiations into the program first, with heavy comments to explain what is going on.  I do not have any functionally useful example that uses these 4 instantiations in the same program, but I can at least use one of them in some made-up sample code.

examples/pna-example-exactmap-ternarymap.p4
```diff
@@ -0,0 +1,319 @@
...
+            // The key fields are all of the fields of struct
+            // tmap1_key.
+            f1: ternary;
+            f2: ternary;
+        }
+        actions = { lookup; }
+        default_action = lookup(42);
+    }
+*/
+    TernaryMap<tmap1_key, bit<16>>(
```
jfingerh, line 185: Add explicit comment that earlier const_entries members are high matching priority than later ones, just like for 'const entries' in P4 tables.

examples/pna-example-exactmap-ternarymap.p4
```diff
@@ -0,0 +1,319 @@
...
+    // The action parameters of action 'lookup' correspond to the
+    // fields of struct 'emap2_val', "exploded" out in order to use
+    // the types and names of the member fields of the struct.
+    action lookup (bit<4> val1, bit<12> val2) {
+        // body of action is irrelevant for control plane API generation
+    }
+    table emap2 {
+        key = {
+            // The key fields are all of the fields of struct
+            // emap2_key.
```
jfingerh, line 142: Bug in `table emap2` key fields.  Fix it!

jfingerh: fixed.

examples/pna-example-exactmap-ternarymap.p4
jfingerh: Added such a comment in definition of const_entries for TernaryMap externin pna.p4 file.

pna.p4
```diff
@@ -545,6 +545,230 @@ extern Digest<T> {
...
 // END:Digest_extern
 
+// BEGIN:ExactMap_extern
+struct exactmap_const_entry_t<K,V> {
+    K key;
+    V val;
+}
+
+struct exactmap_initial_entry_t<K,V> {
+    bool const_entry;
```
det-intel, line 555: Adding this bool "const" member to the struct complicates the already challenging semantics of const. It would be worth documenting a bit more clearly what it means. Does it mean it can't be deleted by any runtime API? What about automatic add/del ops? Does it change the table props in any other way (like current const does on tables).

jfingerh: A short answer is that `const entries` in P4 tables means all entries are const, and none can be removed, modified, NOR added (this is in the language spec).  The constructor with `const_entries` is for these externs is intended to be like that.

The pull request that Andrew Pinski linked to is a proposed new thing in the language spec defining `entries` for a P4 table without a `const` qualifier.  See there for details, but the intent is that each of the entries can be independently marked as `const` or not, and only those marked `const` are prevented from being removed or modified by the control plane, but the control plane is free to add new entries to the table (up to the size limitation).  The constructor for these externs with `initial_entries` parameter is intended to work like that, and the true/false value for `const` takes the place of the `const` keyword.

While it would be nice if the same syntax could be used for P4 tables as here, that would be a bigger change to the spec, e.g. perhaps to define the current syntax in P4 tables as a syntax for a new type of object, perhaps called a 'table entry' object, but I wasn't hoping to go that far here.

jfingerh: If any or all of the text in my previous comment could be usefully added to this PR, let me know.  I think some of it is said in comments in the pna.p4 include file, e.g. it says elsewere in this PR:

const_entries is a list of entries, similar to the 'const
     * entries' table property for tables.  This set of entries cannot
     * be removed or modified by the control plane, and also, the
     * control plane is not allowed to add any entries to an ExactMap
     * instance created using this constructor.

And for initial_entries it is commented as:

The same as the ExactMap constructor with a parameter named
     * `const_entries`, except that the control plane is allowed to
     * add entries to an ExactMap instance constructed using this
     * constructor (subject to capacity constraints, as usual), and
     * the control plane can modify or remove any entries that has a
     * `const_entry` field equal to false.  Any entries with a
     * `const_entry` field value equal to true cannot be modified or
     * removed by the control plane.

examples/pna-example-exactmap-ternarymap.p4
```diff
@@ -0,0 +1,327 @@
...
+        actions = { lookup; }
+        default_action = lookup(42);
+    }
+
+    Note that as of 2022-Nov, the open source p4c front end gives a
+    compile time error if there is a table and an extern instance
+    defined within a control that have the same name as each other.
+    Thus we can generate control plane API files, e.g. P4Info files,
+    assuming that extern instances of ExactMap and TernaryMap will
+    never have the same names as any P4 tables in the same control.
```
jfingerh, line 125: @jamescchoi See the paragraph comment I just added earlier today (2022-Nov-29) following up on a question in a meeting we had earlier today.

jafingerhut:
The 2022-Dec-05 P4 language design work group meeting resulted in the `list` type being added to the P4 language spec, and it is part of open source p4c as I mentioned in an earlier comment, so at least in those ways, there is no impediment to considering this for PNA now.

I am mindful of Thomas Calvert's comment here https://github.com/p4lang/pna/pull/52#pullrequestreview-1189715364   Unfortunately the topic of generalizing P4 tables was not discussed at the 2022-Dec-05 language design work group meeting, at least partly due to a full plate of topics.

@thomascalvert-xlnx  If you would like that issue discussed at the next (2nd Monday of January) language design work group meeting, we may as well suggest adding it to the agenda now.  Would you like me to propose the discussion topic there, or would you like to?

thomascalvert-xlnx:
@jafingerhut thanks for the poke, I've just sent Mihai an email to suggest adding it to the agenda.

jfingerh:
@jafingerhut Add link to P4 LDWG issue related to this topic for reference.

PR #53, opened 2022-08-29
Example: crypto_accelerator object extern

pbhide:
This example demonstrates a definition of a crypto_accelerator
engine as an extern object and its use.

The example assumes that inline accelerators are at the end of
a pipeline and operate on the packet after it is deparsed.
The packet is recirculated after decryption so that the original
packet can be processed again.

On encryption, the packet is sent out after encrypt w/o a need
for recirculation.

The example shows the use of get_results() method in parser()
as well as MainControl()


examples/crypto-accelerator.p4
```diff
@@ -0,0 +1,508 @@
...
+    void set_icv_offset<T>(in T offset);
+    void set_icv_len<T>(in T len);
+
+    /// Setup payload to be encrypted/decrypted
+    void set_payload_offset<T>(in T offset);
+    void set_payload_len<T>(in T len);
+
+    /// Operation
+    void encrypt<T>(in T enable_auth);
+    void decrypt<T>(in T enable_auth);
```
thomascalvert-xlnx, line 73: Following today's discussion, perhaps these should explicitly say these operations are asynchronous, e.g. `mark_for_encrypt`, `encrypt_async`, `prime_encryption`, etc.

pbhide: All methods are async as the assumption as stated in the PR's description is that the crypto accelerator is at the end of the pipeline. Since all methods are async, I think it is not necessary to name them as async. This also seems to be the convention with other externs such as drop_packet() or send_to_port() These actions take effect after completing the execution of the current pipeline (including deparser).
See description of drop_packet here - 
https://github.com/p4lang/pna/blob/main/pna.p4#L642

I can add more description to this effect in the comments.. but prefer not to change the names.
When we add sync method we should name it as such since that does not fit current architecture.

pbhide: changed the names as encrypt -> enable_encrypt, decrypt -> enable_decrypt and also added some comments to indicate that this operation happens after deparser.

jfingerh:
@pbhide Would it be reasonable to give a specific example of the behavior of encryption, by showing an example packet after it has been deparsed, but before the decrypt block begins processing the packet, and then also the modified contents of the packet just after the decrypt block finishes processing the packet?

Similarly for an encrypted packet.

jfingerh:
@pbhide For IPSec decryption, would your proposal include the behavior that the decrypt block does the anti-replay attack detection check?  Or would you propose that such a check must be done explicitly in P4 code somehow, before sending the packet to the decryption block?

jfingerh:
@pbhide For decryption of packets that also have an message digest authentication, e.g. an IPsec AH header, does your proposal include that the decryption block would check the contents of this AH header, and one of the possible error conditions during decryption would be a failure of the authentication?

pbhide:
> @pbhide Would it be reasonable to give a specific example of the behavior of encryption, by showing an example packet after it has been deparsed, but before the decrypt block begins processing the packet, and then also the modified contents of the packet just after the decrypt block finishes processing the packet?
> 
> Similarly for an encrypted packet.

You mean show packet in the comments? 
From programmers point of view `ipsec_esp_decrypt()` action adds a recirc header to the packet. `set_auth_data_xxx` apis may add some information to packet that is specific to target implementation.. so it cannot be show here. If you follow the parser path on recirc, that indicates all the header that are present after decrypt.

Similarly, if we look at `ipsec_esp_encrypt()` new headers (outer header of a tunnel) are added by this action. ICV and trailer that is specified by AES-GCM related rfc are added by crypto engine if instructed to do so by using `set_icv_xxx` methods.

pbhide:
> @pbhide For IPSec decryption, would your proposal include the behavior that the decrypt block does the anti-replay attack detection check? Or would you propose that such a check must be done explicitly in P4 code somehow, before sending the packet to the decryption block?
> 
> If the decryption block does the anti-replay attack detection, would one of the reasons for failing to decrypt the packet be that this check failed?

SA specific random number and extended sequence numbers are expected to be handled in the P4 code.

jfingerh:
@pbhide wrote "You mean show packet in the comments?"

If there is a more convenient way to show such documentation, e.g. text and a figure in the PNA specification, or a separate document that is in this pna repository, that sounds like it might be much more convenient to write and maintain than trying to do ASCII art in comments.  Whatever is easiest for you.  If it is not in comments, the comments could mention where a P4 developer can go to get more details, e.g. "see section <foo> of document <bar> for more details".

pbhide:
> @pbhide For decryption of packets that also have an message digest authentication, e.g. an IPsec AH header, does your proposal include that the decryption block would check the contents of this AH header, and one of the possible error conditions during decryption would be a failure of the authentication?

This example assume ipsec ESP which provides both authentication and confidentiality, not AH.

jfingerh:
In general, it would be good to be very precise about the "division of labor" for decrypting and encrypting IPsec packets between what you expect the P4 code to do, and what you expect the decryption/encryption block to do, including any packet modifications made by the block on its own, independent of P4 code, whether the decrypt block does anti-replay attack protection check or not, whether it does authentication checks or not, which options of IPsec ESP and/or AH you expect it will support, versus which it will not.

Of course these details might change in future versions, but it seems worth trying to at least give your expected answer to such questions.

pbhide:
> Of course these details might change in future versions, but it seems worth trying to at least give your expected answer to such questions.

I'll add this information in the comments where the extern object is defined to make it clear wrt assumed functionality of the hardware

pbhide:
@jafingerhut I have updated the comments in crypto-accelerator.p4 to provide the description/assumptions about crypto engine functionality. PTAL.

pbhide:
> @pbhide Would it be reasonable to give a specific example of the behavior of encryption, by showing an example packet after it has been deparsed, but before the decrypt block begins processing the packet, and then also the modified contents of the packet just after the decrypt block finishes processing the packet?
> 
> Similarly for an encrypted packet.

Added ESP packet format.. and text to indicate headers that are present before and after decrypt and encrypt

pbhide:
> @pbhide For decryption of packets that also have an message digest authentication, e.g. an IPsec AH header, does your proposal include that the decryption block would check the contents of this AH header, and one of the possible error conditions during decryption would be a failure of the authentication?

This example does not show AH based auth. But it can be done where P4 pipeline include outerIP header (clear the mutable fields etc), AH header when providing the auth data offset. The engine will compute and compare ICV over it.

examples/include/crypto-accelerator.p4
```diff
@@ -0,0 +1,136 @@
...
+}
+
+#define ICV_AFTER_PAYLOAD ((int<32>)-1)
+
+/// The crypto_accelerator engine used in this example uses AES-GCM algorithm.
+/// It is assumed to be agnostic to wire protocols i.e. does not understand protocol 
+/// specific headers like ESP, AH etc
+///
+/// Note that crypto accelerator does not modify the packet outside the payload area and ICV
+///     Any wire-protocol header, trailer add/remove is handled by P4 pipeline
```
jfingerh, line 40: Today there is no standard way for a P4 pipeline to add/remove trailers at the end of long pipelines.  The best workaround for short packets is to completely parse every byte of the short packet, and add/remove headers at the end, but that is not a good solution for packets that can be MTU in size.

Does your solution propose new capabilities in PNA and/or P4 to add/remove trailers?

jfingerh: Alternately, one could avoid the need for handling trailers in P4 pipeline completely by bundling together trailer add/remove into the encrypt/decrypt block logic.

mbudiu-vmw: The architecture can do anything it wants with trailers, that's how ethernet trailers actually work.
But you could add some externs to manipulate them. Symmetric with packet_in, but for trailers.

pbhide: There are multiple ways to add/manipulate trailers. This capability is useful in endpoints (PNA where protocols like ESP, RDMA do use trailers). One method can be - payload is defined as a variable length header. Parser can extract it (based on length computed from ipHeader.totalLen or any other means available). Deparser similarly can add it back to packet and then add trailer headers after it. This will not need any new support in P4 language.
Many targets and/or target compilers allow large extractions as long as control functions do not access the extracted bytes as in this case.

pbhide: For targets that need trailer processing by the accelerator, more methods can be added to this extern object for handling the trailers.. but those will be target specific, so I have not shown them here.

pbhide: Accessing/using trailers is done by more than one protocols so it is possible to define target-specific externs to add/remove/modify trailers that are generically useful outside IPSec ESP. Since the inline accelerators are after deparser, it is expected that trailers are added *before* packet reaches the accelerator. For crypto accelerators, trailers may be just payload that will be encrypted.
From the division of work perspective, I did not want to include this capability in the accelerator. In case it is required for specific targets, we can define target-specific methods for it as mentioned in earlier comment.

examples/include/crypto-accelerator.p4
```diff
@@ -0,0 +1,136 @@
...
+/// Crypto accelerator Extern
+enum bit<8> crypto_algorithm_e {
+    AES_GCM = 1
+}
+enum bit<8> crypto_results_e {
+    SUCCESS = 0,
+    AUTH_FAILURE = 1,
+    HW_ERROR = 2
+}
+
```
pbhide, line 26: Add comments here

examples/include/crypto-accelerator.p4
```diff
@@ -0,0 +1,136 @@
...
+    // disable crypto engine. Between enable and disable methods,
+    // whichever method is called last overrides the previous calls
+    void disable();
+
+    crypto_results_e get_results();       // get results of the previous operation
+
+    // set_error_action() indicates behavior on encoutering an error
+    // Default action is NO_ACTION i.e. report error via get_results()
+    // Certain actions may need action parameters, e.g send_to_port
+    void set_error_action<T>(crypto_error_action_e error_action, in T action_param);
```
pbhide, line 135: call it drop_on_error() remove other options. Name set_error_behavior().. or on_error() ? crypto_error_action_e -> crypto_error_behavior_e

examples/include/crypto-accelerator.p4
```diff
@@ -0,0 +1,136 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+/// Crypto accelerator Extern
+enum bit<8> crypto_algorithm_e {
```
apinski-cavium, line 18: Should these not be serializable enums?

examples/include/crypto-accelerator.p4
pbhide: add reciculate as an option

examples/include/crypto-accelerator.p4
pbhide, line 18: make it non-serializable enums.

examples/ipsec-acc.p4
```diff
@@ -0,0 +1,484 @@
...
+        // Remove protocol specific header (E.g. RFC4106)
+        bit<16> encr_pyld_len = hdr.ipv4_2.totalLen;
+
+        ipsec_acc.set_payload_len(encr_pyld_len);
+
+        // TODO: compute padding, build esp_trailer etc.
+
+        // instruct engine to add icv after encrypted payload
+        ipsec_acc.set_icv_offset(ICV_AFTER_PAYLOAD);
+        ipsec_acc.set_icv_len((bit<32>)4); // Four bytes of ICV value.
```
apinski-cavium, line 366: Minor issue:
32w4 instead of the cast.

jfingerh:
@pbhide Sorry, a couple of more questions:

For replay attack detection, are you thinking that this would be done in P4 code _after_ the decryption block and packet recirculation?  If so, that makes sense, and I would recommend that you mention this somewhere.  If you meant before, then that also means before authentication, which would leave you open to attackers that, without knowing your crypto details at all, can make you update your replay attack detection state for packets never sent by the true sender, i.e. a cheap denial of service attack.

Are you planning to create a P4 program example that uses this extern, demonstrating a complete packet decryption code, with all of the things that the P4 should do before and after the decryption block is used?  And similarly for packet encryption code?  I think those would be very valuable examples, even if they do not demonstrate all of the options.

pbhide:
> @pbhide Sorry, a couple of more questions:
> 
> For replay attack detection, are you thinking that this would be done in P4 code _after_ the decryption block and packet recirculation? If so, that makes sense, and I would recommend that you mention this somewhere. If you meant before, then that also means before authentication, which would leave you open to attackers that, without knowing your crypto details at all, can make you update your replay attack detection state for packets never sent by the true sender, i.e. a cheap denial of service attack.
> 
> Are you planning to create a P4 program example that uses this extern, demonstrating a complete packet decryption code, with all of the things that the P4 should do before and after the decryption block is used? And similarly for packet encryption code? I think those would be very valuable examples, even if they do not demonstrate all of the options.

I have mentioned that anti-replay checks are not shown in this example... This is not a complete ipsec example wrt P4 pipeline functions.. this has a few short-cuts to just show the usage of the extern. I don't plan to provide a complete working ipsec example here as this proposal is not yet accepted. It it gets accepted, we may invest in it.. but as such it is not necessary to demonstrate the use of the extern.

examples/ipsec-acc.p4
pbhide: done

examples/include/crypto-accelerator.p4
pbhide: done

jfingerh:
@pbhide wrote: "I have mentioned that anti-replay checks are not shown in this example... This is not a complete ipsec example wrt P4 pipeline functions.. this has a few short-cuts to just show the usage of the extern. I don't plan to provide a complete working ipsec example here as this proposal is not yet accepted. It it gets accepted, we may invest in it.. but as such it is not necessary to demonstrate the use of the extern."

Makes sense.  I will create a separate issue to track the creation of an example P4 program that hopefully demonstrates all of the pieces of a working IPsec decryption case, plus IPsec encryption case, using the extern, once it is merged in.

examples/include/crypto-accelerator.p4
pbhide: This method was added as a result of discussion to see if it makes sense. I am removing it for the following reasons - 
1. As mentioned in the code comments, it is preferable to keep the accelerator agnostic to packet redirection (send to  port), packet drop type of functions that are not related to crypto operation. 
2. Also many existing targets do not support this functionality. 
3. Any targets existing or future that support it, can add a target specific method for this.

examples/include/crypto-accelerator.p4
pbhide: done

jfingerh (approved):
While we may want to edit this before Christmas 2022, I think this is in a fine state to merge in as it is now, and fine-tune it with future small PRs as desired.  Will discuss at 2022-Oct-31 PNA meeting to see if there is any objection.  If no big objections, I hope to merge this as it is on its 7th commit.

thantry:
LGTM

jafingerhut:
We have a couple of approvals on this, and no objections mentioned, so merging it in.  Feel free to create PRs for review with modifications to this, if desired, and they can be considered.

PR #59, opened 2022-10-03
Example: crypto inline method addition to crypto_accelerator object

loalan:
This example expands on the definition of crypto_accelerator (https://github.com/p4lang/pna/pull/53).

The example adds two methods for encrypt/decrypt that assumes that inline accelerators operate immediately on the packet (e.g. deparse, decrypt and reparse).  Packet recirculation is not necessary for either inline method.

The example shows the use of inline encrypt and decrypt, as well as how the crypto accelerator results can be used.


pbhide:
We should separate out hidden/implicit side effects such as 'reparse' and associated state modification from crypto.encrypt/decrypt methods and generalize it so that it can be used for other accelerators as well.
Consider the following -
```
action ipsec_esp_decrypt(...) {
    ....
   <existing methods to setup the decryption parameters>
    ....
    ipsec_acc.decrypt(enable_auth);

    if (reparse_packet() == True) {
        results = ipsec_acc.get_results()
        .... <check results, post_decrypt_actions >....
    } else {
        recirc_packet();
        exit();
    }
}
```
I see the following advantages of this -
1. We can define/debate reparse_packet() extern independent of accelerator methods.
2. This will be useful for all accelerators (present and future).
3. Programmer will be able launch multiple accelerators and reparse the packet once. Mixing this with accelerator method will prevent that.

Overall I think it will keep accelerator functions clean and code slightly more portable across targets.

roop-nvda:
IMHO: This is indeed good example for what we may expect to see with inline accelerators, but there are other (than IPSec) inline accelerations that will need to be expressed as well. The value of inline acceleration in this domain is clear.  Also, most not-trivial accelerators will likely change a multitude of things. My hope is that the solution we land of is general enough to be reused in other inline accelerators that will be needed in the near future by all vendors.

examples/crypto-inline.p4
```diff
@@ -0,0 +1,102 @@
...
+    void enable_encrypt<T>(in T enable_auth);
+    void enable_decrypt<T>(in T enable_auth);
+
+    // encrypt_inline runs immediately and returns control flow to the current pipeline
+    // stage. The method is responsible for encrypting the payload appropriately, creating
+    // the ESP header, calculating the payload offset and lengths, and reparsing the packet
+    void encrypt_inline(in crypto_mode_e mode,
+                        in bit<1> enable_auth,
+                        in bit<32> spi,
+                        in bit<32> seq,
```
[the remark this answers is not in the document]
loalan: Updated with template types

examples/crypto-inline.p4
```diff
@@ -0,0 +1,102 @@
...
+                        in bit<64> iv);
```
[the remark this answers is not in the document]
loalan: Updated with template types. Do you feel SPI needs it as well, I added enable_auth, seq, iv to match the other methods use of templates

examples/crypto-inline.p4
```diff
@@ -0,0 +1,102 @@
...
+    AES_GCM = 1
+}
+enum bit<8> crypto_results_e {
+    SUCCESS = 0,
+    AUTH_FAILURE = 1,
+    HW_ERROR = 2
+}
+
+enum bit<2> crypto_mode_e {
+    TUNNEL = 0,
```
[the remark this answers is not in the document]
loalan: I suppose the difference here is in the async model encrypt/decrypt is in the HW block, and encap/decap is in P4.
In the inline model both encap/encrypt and decrypt/decap are in the HW block. This is because both deparsing and encryption are triggered inline, prior to the P4 deparse stage. Similar for decrypt and removal of the IPSec headers, this executes inline prior to P4 deparser. (Subsequently there is no need to reinject in order check results and to run it through the deparser).

This is also somewhat different from existing P4 implementations, but seem to be what is needed to make this packet transformation inline.

examples/ipsec-inline.p4
```diff
@@ -0,0 +1,376 @@
...
+            ipsec_esp_decrypt(spi, salt, key, key_size, ext_esn_en, auth_en, esn);
+        }
+    }
+    // setup crypto accelerator for decryption - get info from sa table
+    // lookup sa_table using esp.spi
+    table ipsec_sa {
+        key = {
+            // For encrypt case get sa_idx from parser
+            // for decrypt case esp hdr spi value will be used as sa_idx
+            main_meta.sa_index  : exact;
```
rst0git, line 329: Would it make sense to have separate match-action tables for encryption and decryption?

For example, encrypt table could match on the destination IP address, while decrypt table could match on source and destination IP address, as well as SPI value. (This is similar to the approach proposed in [P4-IPsec](https://ieeexplore.ieee.org/document/9151942))

loalan: Sure, that's possible. This is just an example, to focus on how the inline model works

examples/include/crypto-accelerator.p4
```diff
@@ -122,6 +128,29 @@ extern crypto_accelerator {
     void enable_encrypt<T>(in T enable_auth);
     void enable_decrypt<T>(in T enable_auth);
 
+    // crypto accelerator runs immediately and returns control flow to the current pipeline
+    // stage. The method is responsible for creating the ESP header, calculating the payload 
+    // offset and lengths, and then encrypting the payload appropriately and reparsing the 
+    // packet.
+    // H - inout Headers that is the output of the parser block
+    // M - inout Metadata that is from the parser block and shared with the control 
+    crypto_results_e encrypt_inline<H,M,T,S,I>(inout H hdr, inout M meta,
```
loalan, line 137: add a comment of order of operations. e.g. deparser or byte stream update must occur when the extern is invoked

loalan: add comment on preconditions and postconditions

loalan: add side effects

loalan: it may need packet in as well

examples/ipsec-inline.p4
loalan: Actually, what you proposed is simpler than setting parser metadata. I'll look at rewriting the example app.

apinski-cavium:
https://p4.org/p4-spec/docs/P4-16-working-spec.html#sec-packet-data-extraction
```
The packet_in extern is special: it cannot be instantiated by the user explicitly. Instead, the architecture supplies a separate instance for each packet_in argument to a parser instantiation.
```

examples/include/crypto-accelerator.p4
```diff
@@ -122,6 +128,55 @@ extern crypto_accelerator {
...
+    // S - in seq is the optional sequence number
+    // I - in iv is the initialization vector
+    crypto_results_e encrypt_inline<H,M,T,S,I>(packet_in pkt,
+                        inout H hdr, 
+                        inout M meta,
+                        in crypto_mode_e mode,
+                        in T enable_auth,
+                        in bit<32> spi,
+                        in S seq,
+                        in I iv);
```
thomascalvert-xlnx, line 155: Why include all these arguments rather than reuse the `set_iv()`, `set_key()`, etc. methods above?

examples/include/crypto-accelerator.p4
```diff
@@ -122,6 +128,55 @@ extern crypto_accelerator {
     void enable_encrypt<T>(in T enable_auth);
     void enable_decrypt<T>(in T enable_auth);
```
thomascalvert-xlnx, line 130: Do we expect any targets to support both `enable_encrypt` and `encrypt_inline` ? If not, how would a program be written to be portable?

thomascalvert-xlnx (commented):
I suspect that the programming model proposed here (pseudo-recirculation) will be a source of confusion and bugs. For example, pre-conditions in the control block (leading up to the `decrypt_inline` call) could change, resulting in broken programmer's assumptions as well as inability to perform static analysis. Most targets are likely to implement this functionality by recirculating anyway, so it's not clear to me what we are gaining here with this clunky passing of pkt/hdr/meta arguments (imho).

The same effect can be easily achieved explicitly by the programmer:
```
control {
  if (recirculating == false) {
    (...) // set things up
    result = decrypt_inline(...);
    if (result == SUCCESS) {
      recirc(); // or whatever
    } else {
      (...) // deal with error
    }
  }
  else {
    (...) // continue processing decrypted packet
  }
}
```
With the advantage that it can then be used for other purposes too.

PR #60, opened 2022-10-04
Add command line option to p4test to allow 'if' statements in P4 actions

jfingerh:


examples/runp4test.sh
```diff
@@ -1,8 +1,22 @@
...
+# There are some PNA programs that we want to write, and check their syntax
+# using p4test, that have such 'if' statements.
+
+# This proposed PR for p4c would enable such programs to be compiled
+# without error in a different way, but does not work yet as of 2022-Oct-04:
+# https://github.com/p4lang/p4c/pull/3549
+
+P4TEST_OPTS="--excludeMidendPasses Predication"
+
 for j in *.p4
```
pbhide, line 17: I did not know about this script.. but I split my ipsec example into two files.. one file only has the definition of the extern which is included in the main example.. I am not sure if the file with extern def only will compile on it own.

Should we create an include dir inside example dir and move 'included' files there?

pbhide (approved):

PR #62, opened 2022-10-21
Complete match_kinds TBD

hesingh:
See https://github.com/p4lang/p4c/pull/3588

### DIFF
--- a/PNA.mdk
+++ b/PNA.mdk
@@ -514,11 +514,8 @@ these statements is true:
 ~ End P4Example
 
 ## Match kinds
-~Begin TBD
-Consider simply referencing the corresponding section of the PSA
-specification for this, unless we want to have something different in
-PNA.
-~End TBD
+PNA supports the `match_kinds` specified in section 4.3 of the
+PSA specification.
 
 ## Data plane vs. control plane data representations {#sec-data-plane-vs-control-plane-values}
 
--- a/pna.p4
+++ b/pna.p4
@@ -336,7 +336,8 @@ enum PNA_IdleTimeout_t {
 // BEGIN:Match_kinds
 match_kind {
     range,   /// Used to represent min..max intervals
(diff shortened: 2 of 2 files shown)


jafingerhut (approved):
LGTM, but I will add this as a discussion topic for 2022-Oct-24 public PNA meeting to see if there are any objections.  It would be good to add `optional` to the pna.p4 include file in this same PR, if you do not mind making that change as well.

thantry:
LGTM

PNA.mdk
```diff
@@ -514,11 +514,8 @@ these statements is true:
 ~ End P4Example
 
 ## Match kinds
-~Begin TBD
-Consider simply referencing the corresponding section of the PSA
-specification for this, unless we want to have something different in
-PNA.
-~End TBD
+PNA supports the `match_kinds` specified in section 4.3 of the
+PSA specification.
```
det-intel, line 518: The PSA spec doesn't include "optional" as a match kind, which is added in this PR. Should we put an explanation of it here? At least note its existence?

apinski-cavium: The PSA issue is https://github.com/p4lang/p4-spec/issues/1089

apinski-cavium: And was added to the PSA spec with https://github.com/p4lang/p4-spec/commit/3676a3c218e5efd3b32560e546084589cd6b1c64 .

jfingerh: but unfortunately not part of a released version of the PSA specification, but that we can fix by Christmas 2022, by releasing updated PSA and PNA specs together.

jafingerhut: I will see if I can get the "working draft" HTML/PDF links on the specs page to show the actual latest version of PSA and PNA specs (and all of them, really): https://p4.org/specs/

PR #70, opened 2022-10-29
Add hash algorithms to pna.p4

hesingh:
Currently hash algorithms are TODO in PNA specification.  This PR attempts to complete the TODO and also add defines to use as a flag to hash api to use hash on only source, destination, or both IP/IPv6 addresses, and protocol.
 
One has to start somewhere and thus psa.p4 hash algorithms have been copied to pna.p4. Also, none of the hash algorithms look like one a Nic can't support.

### DIFF
--- a/pna.p4
+++ b/pna.p4
@@ -343,7 +343,13 @@ match_kind {
 
 // BEGIN:Hash_algorithms
 enum PNA_HashAlgorithm_t {
-  // TBD what this type's values will be for PNA
+  IDENTITY,
+  CRC32,
+  CRC32_CUSTOM,
+  CRC16,
+  CRC16_CUSTOM,
+  ONES_COMPLEMENT16,  /// One's complement 16-bit sum used for IPv4 headers,
+                      /// TCP, and UDP.
   TARGET_DEFAULT      /// target implementation defined
 }
 // END:Hash_algorithms


hesingh:
If folks would like an explicit flag to be provided to a hash, we could use the defines below or put them in an enum and add another hash api with such a flag.

```c
#define IP_FLOW_HASH_DST_ADDR (1<<1)
#define IP_FLOW_HASH_PROTO (1<<2)
#define IP_FLOW_HASH_SRC_PORT (1<<3)
#define IP_FLOW_HASH_DST_PORT (1<<4)
#define IP_FLOW_HASH_REVERSE_SRC_DST (1<<5)    // symmetric hash

```

thomascalvert-xlnx:
> @thomascalvert-xlnx Please do recommend specific text that you think would be good in the commit message for @hesingh, if you have ideas.

The current commit message is simply: "Add hash". I suggested using the PR title, "Add hash algorithms to pna.p4".

hesingh:
I don't think a commit message can be changed after commit.

jfingerh:
@hesingh If not, then a new PR can be created with a different commit message.

thomascalvert-xlnx:
Too late now, but `git commit --amend` is the most convenient way to change an existing commit's message.

https://www.atlassian.com/git/tutorials/rewriting-history

https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History

PR #73, opened 2022-10-31
Rewritable table entry example: TCP connection state tracking - full and simplified

mariobaldi:
Example of how rewritable table entries can be used to implement state tracking of TCP connections. 
A full example code with creation of entries in a TCP connection table using the add_on_entry PNA feature, as well as a simplified example focusing exclusively on rewriteable entries are provided.

### DIFF
--- a/examples/pna-example-tcp-connection-state-tracking-simplified.p4
+++ b/examples/pna-example-tcp-connection-state-tracking-simplified.p4
@@ -129,14 +129,6 @@ parser MainParserImpl(
     }
 }
 
-struct ct_tcp_table_hit_params_t {
-    bit<32> n2h_seqNo;
-    bit<32> h2n_seqNo;
-    bit<32> n2h_ackNo;
-    bit<32> h2n_ackNo;
-    // other connection state being tracked can be added here
-}
-
 control MainControlImpl(
     inout headers_t  hdr,
     inout metadata_t meta,
--- a/examples/pna-example-tcp-connection-state-tracking-simplified.p4
+++ b/examples/pna-example-tcp-connection-state-tracking-simplified.p4
@@ -1,0 +1,249 @@
+Copyright 2022 AMD
+/*
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#include <core.p4>
+#include "../pna.p4"
+
+
+// Sample PNA program demonstrating the use of write back table
+// entries. Although a possible syntax is proposed to indicate
+// action data that is written back to a table, the purpose of
+// this example is not to support that specific syntax, but
+// rather the usefulness and advantages of being able to write
+// action data back to a table to store state information
+
+// NOTE: this example uses a simplified version of the code in
+// pna-example-tcp-connection-tracking.p4 to add entries
+// in the TCP connection table (ct_tcp_table).
+// Although the same naming is purposedly used for the connection
+// table, the removal of entries (due to time out or explicit connection
+// tear-down is explicitly not supported for the sake of keeping the code
+// simpler and more legible.
+// A complete implementation of TCP connection and status tracking
+// would require to include such functionalities that can be found in
+//  pna-example-tcp-connection-tracking.p4
+
+
+// NOTE: the code is currently written for a compiler supporting
+// if statements within actions. The same functionality could be rewritten
+// for a compiler not supporting if statements within actions by using if
+// if statements in the control block and a table to set up proper context.
+
+
+header ethernet_t {
+    bit<48> dstAddr;
+    bit<48> srcAddr;
+    bit<16> etherType;
+}
+
+header ipv4_t {
+    bit<4>  version;
+    bit<4>  ihl;
+    bit<8>  diffserv;
+    bit<16> totalLength;
+    bit<16> identification;
+    bit<3>  flags;
+    bit<13> fragOffset;
+    bit<8>  ttl;
+    bit<8>  protocol;
+    bit<16> hdrChecksum;
+    bit<32> srcAddr;
+    bit<32> dstAddr;
+}
+
+header tcp_t {
+    bit<16> srcPort;
+    bit<16> dstPort;
+    bit<32> seqNo;
+    bit<32> ackNo;
+    bit<4>  dataOffset;
+    bit<4>  res;
+    bit<8>  flags;
+    bit<16> window;
+    bit<16> checksum;
+    bit<16> urgentPtr;
+}
+
+// Masks of the bit positions of some bit flags within the TCP flags
+// field.
+const bit<8> TCP_ACK_MASK = 0x10;
+const bit<8> TCP_SYN_MASK = 0x02;
+
+// Define names for different expire time profile id values.
+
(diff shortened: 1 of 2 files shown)


examples/pna-example-tcp-connection-state-tracking-simplified.p4
```diff
@@ -0,0 +1,249 @@
...
+            default: accept;
+        }
+    }
+    state parse_tcp {
+        pkt.extract(hdr.tcp);
+        transition accept;
+    }
+}
+
+struct ct_tcp_table_hit_params_t {
```
pbhide, line 132: if this struct is used as action parameter, can you show how to specify inout property on individual fields?
Will it be on entire struct? We need ability to specify this per member of the struct.

examples/pna-example-tcp-connection-state-tracking-simplified.p4
```diff
@@ -0,0 +1,249 @@
...
+            @tableonly   ct_tcp_table_hit;
+            @defaultonly ct_tcp_table_miss;
+        }
+
+        add_on_miss = true;
+
+        default_idle_timeout_for_data_plane_added_entries = 1;
+
+        idle_timeout_with_auto_delete = true;
+        const default_action = ct_tcp_table_miss;
```
pbhide, line 220: how do we specify 'lock' property on the table so that any concurrent access to the table are serialized to ensure data integrity? Can we add a property on the table? Or is it expected to be implicitly handled by compiler? What if programmer does not require locking?

examples/pna-example-tcp-connection-state-tracking-simplified.p4
```diff
@@ -0,0 +1,249 @@
...
+// Although the same naming is purposedly used for the connection
+// table, the removal of entries (due to time out or explicit connection
+// tear-down is explicitly not supported for the sake of keeping the code
+// simpler and more legible.
+// A complete implementation of TCP connection and status tracking
+// would require to include such functionalities that can be found in
+//  pna-example-tcp-connection-tracking.p4
+
+
+// NOTE: the code is currently written for a compiler supporting
```
pbhide, line 40: no need for this comment. if/else are allowed in action routine by the language.

pbhide (changes requested):
Do we expect all types of table action parameters to be writable (e.g. TCAM tables...)?

examples/pna-example-tcp-connection-state-tracking-simplified.p4
jfingerh: As discussed during the meeting, I think per-action-parameter `rmw` direction is much more easily achieved than doing it on a per-struct-member basis.

Also, this particular struct is very specific to the definition of `add_entry`.  Maybe there is a reasonable way in P4 to define `add_entry` so that it can take a variable number of parameters, varying with the action that one is adding, but the creation of a struct like this for each such `add_entry` call on a different action name is unfortunately part of the current proposal for `add_entry`.  A bit klunky, I know, and if you have better suggestions that fit within the existing P4 language spec, or minor extensions to it, let me know.

examples/pna-example-tcp-connection-state-tracking-simplified.p4
jfingerh: I think a fairly straightforward way to do this is to define this feature so that the program must behave AS IF there were a per-table-entry lock that is acquired when the action begins executing, and is released when the action is finished executing.

Implementers might find other actual implementation methods that achieve the same behavior, and the spec should not restrict those, as long as the behavior is as described above.

examples/pna-example-tcp-connection-state-tracking-simplified.p4
jfingerh: There was a whole big recent discussion on whether add_entry_if should be added to PNA or not, and if so, why, related to some P4 targets not supporting if statements within action bodies.  I hope you are right that P4 implementations for NICs that do not support if within actions will be rare, but they might still exist for a while.

jfingerh:
Thomas Calvert asked a question about: since this proposed new feature could be used to implement counters, is there a reason to keep Counter and DirectCounter externs?

I definitely think those externs should be kept, because they provide "update only" semantics from the data plane APIs, i.e. the data plane is not allowed to read the values of such extern-based counters, and this is important in enabling cheaper implementations of such counters.

I would definitely encourage anyone interested to remind us to add an appendix/footnote/paragraph/whatever to remind P4 developers of this in the PNA spec.

One simple example: fast path data plane might only represent the least significant 32 bits of a 64-bit packet/byte counter in the data plane, and periodically atomically read&clear data plane counters in a driver, and add the values to 64-bit packet/byte counters running on a hefty CPU like a Xeon nearby with lots of cheap DRAM available.  Such an implementation would NOT have the full counter available in the data plane most of the time, but give correct values to control plane APIs for reading counters.

thomascalvert-xlnx:
I think the first 2 lines of each file should be swapped to fix the build. Otherwise LGTM. 👍

jafingerhut:
@mariobaldi I spent a little time editing your 2 example programs so that `p4test` compiles them with 0 error messages using the latest version of p4c code, and the latest version of the pna.p4 include file in this repo.  The modified files are in the attached progs.zip file on this comment.

I removed the `rmw` keyword from the action that had them.  I could #ifdef those if you prefer.

If you would like me or someone else to create a new PR with these versions of the programs, let us know.
[progs.zip](https://github.com/p4lang/pna/files/11024153/progs.zip)

jafingerhut:
@pbhide asked: "Do we expect all types of table action parameters to be writable (e.g. TCAM tables...)?"

My personal answer: A target supports what the target developers choose to enable it to support.  Some targets might enable this, and others might not.

Ideally a target's P4 compiler would reject a P4 program with a compile-time error if any part of it is not supported by that target/compiler, with an error message explaining why.  That is up to the target's P4 compiler back end to check.

mariobaldi:
> @mariobaldi I spent a little time editing your 2 example programs so that `p4test` compiles them with 0 error messages using the latest version of p4c code, and the latest version of the pna.p4 include file in this repo. The modified files are in the attached progs.zip file on this comment.
> 
> I removed the `rmw` keyword from the action that had them. I could #ifdef those if you prefer.
> 
> If you would like me or someone else to create a new PR with these versions of the programs, let us know. [progs.zip](https://github.com/p4lang/pna/files/11024153/progs.zip)

Thanks for doing this @jfingerh. I don't quite get why we don't just commit the changes to this PR, rather than creating a new one?

mariobaldi:
> @mariobaldi In your program pna-example-tcp-connection-state-tracking.p4, the ct_tcp_table_miss action assigns an initial value for either the parameter `n2h_seqNo`, or the parameter `h2n_seqNo`, but never for both of those, and never assigns an initial value for the action parameters `n2h_ackNo` or `h2n_ackNo`
> 
> Thus if the first packet to execute action `ct_tcp_table_hit` for a flow has SYN=0, that action will read uninitialized values of `n2h_ackNo` or `h2n_ackNo`.
> 
> I am sure my earlier TCP connection tracking example is also full of such possible behavior, and I added a comment to it trying to make it explicit that it was primarily a demo of the `add_on_miss` table property and `add_entry` extern function call.
> 
> Would you consider these examples similarly as examples of using assignable action parameters that maintain their values across packets, which might work usefully if packets arrive in the expected sequence, but anyone wanting a production-quality tested version must write their own? If so, makes perfect sense to me.

Yes, you are right. I would consider these just as sample examples. In normal conditions the next packet hitting the entry should have a SYN=1.

mariobaldi:
> @mariobaldi It would be good to put these in a separate directory for now, until the p4test front end compiles them without errors, to avoid Github CI issues.

I guess we don't need it if we commit in this PR @jafingerhut changes in the zip files. Otherwise, I will move them.

jfingerh:
@mariobaldi If you want to take the edited programs in the zip file I attached to an earlier comment, and update this PR to use them, then they should compile cleanly with the latest version of open source p4test, and thus no problems with CI if the programs are in the example directory, and no reason to create a separate PR.

Issue #19, opened 2021-05-15
Should Hash extern for PNA add Toeplitz hash option?


jafingerhut:
This is sometimes specified as a required hash function for implementing Receive Side Scaling by some NIC customers/driver developers.

jfingerh:
@thantry This is an issue you and I discussed recently, that there is currently nothing said about it in the PNA spec.  No details here yet, just a reminder for future discussions.

jfingerh:
@thantry This would be a good issue to add links to details about Toeplitz hash added to p4lang/behavioral-model that you mentioned in 2022-Nov-14 meeting.

thantry:
Here's the implementation @qobilidop added to the BMv2 backend.

https://github.com/p4lang/behavioral-model/pull/1132
https://github.com/p4lang/behavioral-model/pull/1134

qobilidop:
I can provide some additional context for your reference.

My implementation in https://github.com/p4lang/behavioral-model/pull/1134 was basically following the implementation for [`crc16_custom` and `crc32_custom` hashes declared in v1model.p4](https://github.com/p4lang/p4c/blob/a56268106d7200490ed0049519a29ed327a595b7/p4include/v1model.p4#L394-L403), which also require internal states to be set. I dug into the history a little bit, and found this comment:

https://github.com/p4lang/behavioral-model/issues/982#issuecomment-786353701
> bmv2 supports runtime-configurable CRCs, when selecting `crc32_custom` or `crc16_custom` as the compile-time algorithm. But these are only configurable through the runtime CLI at the moment, not P4Runtime.

So I was wrong when I previously said there is a P4Runtime interface to configure Toeplitz RSS key with bmv2. It's only configurable through the bmv2-specific P4 runtime CLI.

I wonder if we could provide a P4Runtime interface to configure the RSS key for the Toeplitz hash if we declare it as an extern object like the following (adapting from [pna.p4](https://github.com/p4lang/p4c/blob/a56268106d7200490ed0049519a29ed327a595b7/p4include/pna.p4#L280-L300)):

```p4
extern ToeplitzHash<O> {
  Hash();
  void set_rss_key<K>(in K key);
  O get_hash<D>(in D data);
  O get_hash<T, D>(in T base, in D data, in T max);
}
```

Then there are probably standard ways to invoke the `set_rss_key()` method (maybe as a table action) through P4Runtime.

jfingerh:
[...]

PR #91, opened 2022-11-16
Add hash algorithms to pna.p4

hesingh:
Currently hash algorithms are TODO in PNA specification. This PR attempts to complete the TODO and also add defines to use as a flag to hash api to use hash on only source, destination, or both IP/IPv6 addresses, and protocol.

One has to start somewhere and thus psa.p4 hash algorithms have been copied to pna.p4. Also, none of the hash algorithms look like one a Nic can't support.

### DIFF
--- a/pna.p4
+++ b/pna.p4
@@ -343,7 +343,13 @@ match_kind {
 
 // BEGIN:Hash_algorithms
 enum PNA_HashAlgorithm_t {
-  // TBD what this type's values will be for PNA
+  IDENTITY,
+  CRC32,
+  CRC32_CUSTOM,
+  CRC16,
+  CRC16_CUSTOM,
+  ONES_COMPLEMENT16,  /// One's complement 16-bit sum used for IPv4 headers,
+                      /// TCP, and UDP.
   TARGET_DEFAULT      /// target implementation defined
 }
 // END:Hash_algorithms


jafingerhut (approved):
Looks reasonable to me.  If a particular NIC cannot support some of these, it is pretty common practice for particular targets to only support the subset that they can, and I expect that to be true of PNA target devices as well.

hesingh:
Please merge, thanks.

jfingerh:
@hesingh Happy to do so after having at least an announcement and call for objections or concerns in the next public PNA work group meeting next Monday.  I do not anticipate any, but I'm trying not to ram things through without such notification.

thantry:
Would be good to add Toeplitz and Xor hash, which are common in many NICs for RSS.

thomascalvert-xlnx:
LGTM.

Agree that Toeplitz needs adding, but equally happy for that to be in a separate commit.

jfingerh:
@thantry @thomascalvert-xlnx Thanks for review.  Yes, we should add other hash algorithms as desired.  For Toeplitz, this issue is tracking that, and from previous conversations it sounds like there might be additional control plane and/or data plane APIs desired there, and if so, merely adding a new enum value to the list is not enough, and I'd rather someone propose all of the pieces necessary in a PR addressing that issue: https://github.com/p4lang/pna/issues/19

For XOR, I have created a separate issue to track that: https://github.com/p4lang/pna/issues/93   I understand the desire to pile all the desired changes into single PRs, but hopefully we can get to a place where separate PRs can be moved through the pipeline of writing, approvals, and merging quickly enough that we are comfortable doing separate PRs.

mariobaldi (approved):

PR #92, opened 2022-11-20
Add description of new value AUTO_DELETE for table property pna_idle_timeout

jfingerh:
Also clarify some wording in the description of add-on-miss, and remove the old signature in the document, in favor of using Madoko INCLUDE mechanism to include the corresponding section of the pna.p4 include file, so that the document and include file will remain in sync with each other.


PNA.mdk
```diff
@@ -902,24 +904,39 @@ A PNA implementation need not support both `pna_implementation` and
...
+lookup results in a miss, `t`'s default action is executed.  So far,
+this is all standard behavior as defined in the P4~16~ language
+specification.
+
+If `t`'s default action makes a call to `add_entry`, it causes a new
+entry to be added to the table with the same key that was just looked
+up and resulted in a miss, and the action name and action parameters
+specified by the parameters of the call to the `add_entry` extern
+function.  Thus, future packets that invoke `t.apply()` with the same
+lookup key will get a match and invoke the specified action (until and
```
thantry, line 922: Does this not assume in-order processing of packets across multiple flows? Should we not state this explicitly?

thomascalvert-xlnx:
LGTM.

(It did leave me wondering whether the `pna_idle_timeout` and `idle_timeout_with_auto_delete` properties could be unified into a single one... (Maybe make `AUTO_DELETE` an additional member of the `PNA_IdleTimeout_t` enum?) But that's not the topic of this commit. :) )

jfingerh:
@thomascalvert-xlnx I like the idea of making AUTO_DELETE a new member of the enum PNA_IdleTimeout_t.  I suppose some targets might even be able to implement a combination of notifying the control plane AND auto-deleting the entry when an entry expires, but no need to add that unless there is a request to do so.

I will go ahead and change this PR with this idea, and see what others reviewers think.

PNA.mdk
jfingerh: I agree that there is a potentially subtle interaction here between pipelines that can have packets "pass each other up" in the middle of the pipeline, versus those that do not, and it would be good to address this issue somehow in the PNA spec.  I will create a separate issue to track this, since I doubt I will quickly find an acceptable description, and that property of a packet processing pipeline probably affects other features besides this one.

jfingerh: Created this issue: https://github.com/p4lang/pna/issues/96  Feel free to add comments on the topic there, but I hope it is reasonable to allow this PR to be merged before that issue is fully resolved.

thomascalvert-xlnx (commented):
Yes I like the AUTO_DELETE option, it looks cleaner IMHO.

Suggest that the text talks about how entries can be explicitly deleted (not via timeout but on seeing e.g. a RST packet), and briefly describe the semantics of `ExpireTimeProfileId_t`. But that could easily be added in a separate change to this one.

thantry:
LGTM

PNA.mdk
thantry: Agree, we should merge this.

thomascalvert-xlnx (approved):

mariobaldi (approved):

mariobaldi (approved):

PR #95, opened 2022-11-25
Remove generated-html directory

jfingerh:
Now that the "Working draft" links on the page https://p4.org/specs page are working again, there is not much reason to keep around that directory any more.


jafingerhut (approved):
There are no changes to the spec in this PR, only removing an older version of the generated HTML and PDF files for the spec.

PR #100, opened 2022-12-11
Flex routing

mariobaldi:
Introduction of more flexible routing out of the packet processing block.


PNA.mdk
```diff
@@ -187,33 +187,41 @@ multiple NIC devices that are conformant to the PNA[^PNAPortability].
...
 serializes the headers back into a packet that can be sent onwards.
 
-After the `MainDeparser`, the packet may either:
-* proceed to the message processing part of the NIC, and then typically on
+After the `Deparser`, the packet is processed by one or more
+inline accelerators. The P4 program executed in the `Control` determines
+whether and how each inline accelerator processes the packet by either setting
+corresponding intrinsic metadata or executing methods of corresponding extern
+object(s).
+Upon existing the inline accelerators, a packet may either:
```
jfingerh, line 217: typo?  existing -> exiting?   Or perhaps another phrase such as "Upon completion of processing in the inline accelerators, ..."

PNA.mdk
```diff
@@ -902,39 +903,24 @@ A PNA implementation need not support both `pna_implementation` and
...
+same key that was just looked up.
+
+The control plane API is still allowed to add, modify, and delete
+entries of such a table, but any entries added via the `add_entry`
+function do not require the control plane software to be involved in
+any way.  It is expected that PNA implementations will be able to
+sustain `add_entry` calls at a large fraction of their line rate, but
+it need not be at the same packet rate supported for processing
+packets that do not call `add_entry`.  The new table entry will be
+matchable when the next packet is processed that applies this table.
```
jfingerh, line 919: I think this text is older than the text it is replacing in this PR?

PNA.mdk
```diff
@@ -902,39 +903,24 @@ A PNA implementation need not support both `pna_implementation` and
...
 
 ~Begin P4Example
-[INCLUDE=pna.p4:add_entry_extern_function]
+extern bool add_entry<T>(string action_name,
+                         in T action_params);
```
jfingerh, line 923: Same here.  The INCLUDE Madoko directive is more recent, and easier to keep up to date with the pna.p4 file if its contents change in the future.

thomascalvert-xlnx:
I'm having trouble reviewing this PR as it seems to duplicate the changes in your previous one -- they are both based on `main`. Should we rebase this one on top of #99 ?

thomascalvert-xlnx:
Should this PR also include changes to the pna.p4 file to reflect the new routing possibilities? The text does not seem to explain how the routing is practically effected, i.e. what code does a programmer need to write in MainControl to send a packet to the host/network?

I am thinking that at least the following structures might need some changes: `PNA_Direction_t`, `PNA_PacketPath_t`, and some of the metadata structures (c.f. comments for `PortId_t` in `pna_main_parser_input_metadata_t`).

mariobaldi:
I am sorry for the confusion. I was undecided as to what was the best way to keep the PR separate from each other so that we could approve or reject different aspects independently. With the goal of making them independent, I based them all on main.

 

---

Mario

mariobaldi:
I kept the text at the same high level it was before (basically saying that the program determines where a packet goes next by setting a piece of metadata or calling an extern function/method.

However, I agree that it would be better to be more specific (and introduce corresponding changes in pna.p4). 

 

---

Mario

PNA.mdk
mariobaldi: Fixed in latest commits

PNA.mdk
mariobaldi: Reverted to newer text

thomascalvert-xlnx (commented):
Text looks good now, thanks Mario 👍

jfingerh (approved):
Reviewing this PR after 5 commits, looks much better than on Monday when it appeared to revert older commits.  LGTM.

thomascalvert-xlnx (approved):
Latest changes look good to me. Does the PDF version get generated correctly?

loalan:
LGTM 👍

PNA.mdk
```diff
@@ -356,15 +324,29 @@ containing the PNA specification[^PNAExamplePrograms].
     in directory `examples`.  Direct link:
     <https://github.com/p4lang/pna/tree/master/examples>
 
+
+## Guidelines for Portability"
```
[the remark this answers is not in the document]
jafingerhut: Good catch.  Will approve and merge this now, and create a small PR later today to address this.

jafingerhut (approved):
LGTM, will create small PR later today to address one comment.

PNA.mdk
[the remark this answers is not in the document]
jafingerhut: Fixed in https://github.com/p4lang/pna/pull/105

PR #101, opened 2022-12-11
Name changes in pna.p4

mariobaldi:
Names changed in pna.p4 to reflect the removal of pre-control.

### DIFF
--- a/pna.p4
+++ b/pna.p4
@@ -574,8 +574,8 @@ enum PNA_PacketPath_t {
     FROM_HOST_RECIRCULATED
 }
 
-struct pna_main_parser_input_metadata_t {
-    // common fields initialized for all packets that are input to main
+struct pna_parser_input_metadata_t {
+    // common fields initialized for all packets that are input to
     // parser, regardless of direction.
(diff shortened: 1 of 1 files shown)


jfingerh (commented):
This might lead to a bit of name churn in current implementations (e.g. DPDK PNA implementation and an internal Intel one), but if we are going to do this, better now than later.

thomascalvert-xlnx (commented):
Makes sense to rationalise the names. Perhaps a single commit changing the names in both the P4 and MDK files would be cleaner, but assuming that both this and #100 get approved around the same time the result will be the same.

This change does make all the examples inconsistent, so I think a mass rename might be in order?

mariobaldi:
I agree that this should go with PR #99, but I thought of keeping them separate in case we agree with changing the architecture (i.e., approve PR #99 and possibly PR#100), but don’t agree with changing the naming convention.

 

---

Mario

mariobaldi:
Obsolete as we decided not to change the names of the parser, deparser and control blocks

Issue #89, opened 2022-11-15
Add text to PSA specification on encrypt/decrypt operations


jfingerh:
While it is definitely valuable to have comments in the definition of the extern, it would also be useful to give some text in the body of the PSA specification, in its own section, describing the basic idea of this extern, when and how its methods should be called, and what their effect is.

@pbhide Is this something you would be willing to write?  Or anyone you can suggest that you can persuade to volunteer for the task?

pbhide:
Will work on it in 1st week of dec.

pbhide:
@jfingerh any recommendation on the section in PNA spec where this should go?
I can think of adding it in section 6 (after 6.10 packet recirculation section, which seems empty btw). OR
I can add it in section 7.
I was looking for a section for "recommended/Required" externs in PNA spec but don't see one. 
Are we going to specify certain extern functions/objects as - "PNA compliant target SHALL support the following extern methods and objects..."

jafingerhut:
I would recommend a new sub-section of section 7, since it is an extern object we are explaining.

PR #103, opened 2022-12-13
Issue #89: Description of Extern Objects for Inline Accelerators

pbhide:
@jafingerhut @thomascalvert-xlnx @mariobaldi 
Since extern objects is one of the ways to represent inline accelerators.. I have added it as an example in the spec.
Let me know if you had something else in mind.

I have *not* done the preview of the .mdk changes.. hopefully they are ok.. I'll work on that while you review the text.

### DIFF
--- a/PNA.mdk
+++ b/PNA.mdk
@@ -864,15 +864,21 @@ done, both the PSA and PNA specifications will reference that.
 
 A variety of inline accelerators can be present on a PNA target. These
 accelerators perform specific functions on a packet. These functions are
-typically implemented in hardware.
+typically implemented in hardware. These accelerators perform specific
+functions on a packet after the deparser has finished executing.
 
 These hardware functions are represented as extern objects in a P4 program.
 An extern object representing a specific accelerator E.g. AES-GCM crypto
-accelerator, can be instantiated in P4 program. The methods defined by
-the extern object are used to control the behavior of the inline accelator
-it represents. This section provides one example definition of a crypto
-acceleration engine. Other extern objects can be defined based on the
-functionality provided by the hardware accelerators.
+accelerator, can be instantiated in a P4 program. The methods defined by
+the extern object are used to send and receive information to/from the
+inline accelerator. Since the accelerators are present after the deparser,
+the information sent to the accelerator takes effect only when packet reaches
+the accelerator. Similarly any information received from accelerator is
+for the previous function performed on the packet.
+
+This section provides one example definition of a crypto acceleration engine.
+Other extern objects can be defined in future based on the functionality
+provided by the hardware accelerators.
 
 ~Begin P4Example
 [INCLUDE=examples/include/crypto-accel.p4:Crypto_accelerator_extern_object]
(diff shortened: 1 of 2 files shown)


jafingerhut:
So perhaps this is too much "angels dancing on the head of a pin" distinction I am making here, but one view is that the extern object you have defined isn't "the accelerator itself" that does the encryption and decryption on the packet, but instead it is merely provides an API for setting parameters that are inputs to the accelerator, and for viewing outputs from the accelerator (e.g. the error status).

I will take a look at the PR soon and add any review comments there.

mariobaldi:
This is also my understanding.

> So perhaps this is too much "angels dancing on the head of a pin" distinction I am making here, but one view is that the extern object you have defined isn't "the accelerator itself" that does the encryption and decryption on the packet, but instead the extern object merely provides an API (via method calls) for setting parameters that are inputs to the accelerator, and for viewing outputs from the accelerator (e.g. the error status).
 
This is also my understanding and I believe the text Parag added to the mdk description reflects this. 

All looks good to me.

mariobaldi (approved):

PNA.mdk
```diff
@@ -860,6 +860,24 @@ extern definitions being moved from the PSA specification into a
 separate standard library of P4 extern definitions, and if this is
 done, both the PSA and PNA specifications will reference that.
 
+## Extern Objects for Inline Accelerators {#sec-extern-restrictions}
+
+A variety of inline accelerators can be present on a PNA target. These
+accelerators perform specific functions on a packet. These functions are
```
thomascalvert-xlnx, line 866: Suggest reminding that the accelerators execute after the pipeline, e.g.:
> These accelerators perform specific functions on a packet, _usually after the `Deparser` has finished executing_.

examples/include/crypto-accelerator.p4
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-/// Crypto accelerator Extern definition
+// BEGIN:Crypto_accelerator_extern_object
```
thomascalvert-xlnx, line 17: This will quote over 100 lines in the document - maybe this marker can be moved to line 82 (just before the `extern` keyword), and readers can come to this source file if they want more details?

(Don't feel strongly about this one.)

thomascalvert-xlnx (approved):
One typo and a couple minor suggestions but overall LGTM.

pbhide:
> 

I have added some description indicating that object methods are used to communicate information to/from the accelerator. Also took @thomascalvert-xlnx suggestion to add/remind about placement of the accelerators. That should clarify the point you are making.

PNA.mdk
pbhide: added

PNA.mdk
```diff
@@ -860,6 +860,24 @@ extern definitions being moved from the PSA specification into a
...
+typically implemented in hardware.
+
+These hardware functions are represented as extern objects in a P4 program.
+An extern object representing a specific accelerator E.g. AES-GCM crypto
+accelerator, can be instantiated in P4 program. The methods defined by
+the extern object are used to control the behavior of the inline accelator
```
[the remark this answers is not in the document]
pbhide: fixed

PR #106, opened 2022-12-20
Some Madoko markup fixes, remove PassNumber types, remove old example…

jfingerh:
… flows


jfingerh:
@mariobaldi I doubt I will do many more changes before a release, but these seems like reasonable changes.  For the example packet flows I am deleting from the appendix, I will probably wait until after the release to create new figures to replace the old ones.

jafingerhut (approved):
LGTM.  These are all either Madoko markup fixes, or removing obsolete figures that could stand to be updated later.

mariobaldi:
The changes look good and I agree with adding the example packet flows after the release

PR #107, opened 2022-12-22
Update comments that referred to old table property names and values

jfingerh:


jafingerhut (approved):
No functional changes to the spec here, only updating obsolete comments so that they are consistent with latest spec.

PR #110, opened 2023-01-09
send_to_port and mirror_packet arguments should be in direction

apinski-cavium:
I noticed this while testing a front-end change for compile time known-constants fix.
These arguments should not be directionless and not required to be a compile time known constants.

Signed-off-by: Andrew Pinski <apinski@marvell.com>

### DIFF
--- a/PNA.mdk
+++ b/PNA.mdk
@@ -574,7 +574,7 @@ each occurs.
 ### Extern function `send_to_port`
 
 ~ Begin P4Example
-extern void send_to_port(PortId_t dest_port);
+extern void send_to_port(in PortId_t dest_port);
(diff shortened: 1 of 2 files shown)


jafingerhut (approved):
LGTM.  Sorry I had not noticed this much earlier!

apinski-cavium:
> LGTM. Sorry I had not noticed this much earlier!

We all missed it before the release in December. Also the examples only use send_to_port/mirror_packet from an action where the argument was from the table/control plane (and p4c has a bug there which allows it) which allowed this to missed by the examples also. Basically I am saying things slip through everyone's eyes even.

PR #111, opened 2023-01-15
Replace 'inline extern' with 'fixed function block'

jfingerh:
The term 'inline extern' is a misleading choice of words for this functionality in PNA, and fixed function block has a longer history of being used for such things.

Remove a left-over mention of the pre control that was missed before.

Correct the name of the extern function set_entry_expire_time_if in some examples in comments.

Use INCLUDE directive in a few more places of PNA.mdk that will help keep the code snippets from the pna.p4 include file consistent between that file and the spec more easily in the future.

### DIFF
--- a/PNA.mdk
+++ b/PNA.mdk
@@ -574,7 +574,7 @@ each occurs.
 ### Extern function `send_to_port`
 
 ~ Begin P4Example
-extern void send_to_port(in PortId_t dest_port);
+[INCLUDE=pna.p4:send_to_port]
 ~ End P4Example
 
 The extern function `send_to_port` is used to direct a packet to a
@@ -597,8 +597,7 @@ without being looped back.
 ## Packet Mirroring
 
 ~ Begin P4Example
-extern void mirror_packet(in MirrorSlotId_t mirror_slot_id,
-                          in MirrorSessionId_t mirror_session_id);
+[INCLUDE=pna.p4:mirror_packet]
 ~ End P4Example
 
 The extern function `mirror_packet` is used to cause a mirror copy of
@@ -786,7 +785,7 @@ the places mentioned in Table [#table-extern-usage].
 |-------------------|----------------------------------------------|
 ~
 
-For example, `Counter` being restricted to "Pre, Main" means that
+For example, `Counter` being restricted to "MainControl" means that
 every `Counter` instance must be instantiated within the `MainControl`
 block, or be a descendant of one of those nodes in the instantiation
 tree. If a `Counter` instance is instantiated in Main, for example, then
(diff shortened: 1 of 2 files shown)


jfingerh:
@mariobaldi These are mostly minor fixes, that I think would be reasonable to merge if you approve, without going through meeting discussions on.

jnfoster (approved):
I have been a proponent of this terminology, so I wholeheartedly approve.

pna.p4
```diff
@@ -793,13 +797,13 @@ extern void update_expire_info(
...
 //
 // Examples:
-// set_expire_time_if(hdr.tcp.flags == TCP_FLG_SYN &&
-//                    meta.direction == OUTBOUND,
-//                    tcp_connection_start_time_profile_id);
-// set_expire_time_if(hdr.tcp.flags == TCP_FLG_ACK,
-//                    tcp_connection_continuation_time_protile_id);
-// set_expire_time_if(hdr.tcp.flags == TCP_FLG_FIN,
-//                    tcp_connection_close_time_profile_id);
+// set_entry_expire_time_if(hdr.tcp.flags == TCP_FLG_SYN &&
```
jnfoster, line 800: Were these changes intentional?

pna.p4
```diff
@@ -793,13 +797,13 @@ extern void update_expire_info(
...
+//                          meta.direction == OUTBOUND,
+//                          tcp_connection_start_time_profile_id);
+// set_entry_expire_time_if(hdr.tcp.flags == TCP_FLG_ACK,
```
jnfoster, line 803: Same.

pna.p4
```diff
@@ -793,13 +797,13 @@ extern void update_expire_info(
...
+//                          tcp_connection_continuation_time_protile_id);
+// set_entry_expire_time_if(hdr.tcp.flags == TCP_FLG_FIN,
```
jnfoster, line 805: Same.

pna.p4
jfingerh: Yes.  The name set_expire_time_if in the comments is not defined in the pna.p4 header file.  set_entry_expire_time_if is defined.

PNA.mdk
```diff
@@ -1057,7 +1056,7 @@ existing slides.
 
 ## Is it inefficient to have the `MainParser` redo work? {#appendix-reparsing}
 
-If the only changes made by the inline extern in the network-to-host
+If the only changes made by the decryption fixed function block in the network-to-host
```
mariobaldi, line 1059: In the architecture picture and description what was previously called "inline extern" is now called "inline accelerators". 
This said, the text also states that "inline accelerators are fixed-function blocks", so even though we use different terms for the same element, the terminology seems coherent to me. But I just wanted to bring this to your attention to make sure you are aware of and fine with it.

mariobaldi (approved):
Thank you for catching these inconsistencies and making these improvements.

PNA.mdk
jafingerhut: Good point.  It seems worth having a short glossary of terms somewhere, along with preferred terms to use in the spec for consistency's sake.  I will merge this one as is, and create a separate issue about the glossary and whether to use "inline accelerator" in most places in the PNA spec.

PR #117, opened 2023-03-20
Issue#108: Remove last remaining occurrences of HOST_TO_NET and NET_TO_HOST from spec.

thomascalvert-xlnx:
Also move the obsolete packet path figures in to the `historical` subdirectory - these will eventually need to be redone as per the TBD in the text (§ 6 and appendix D).

Was able to regenerate the HTML & PDF and the updated table 1 looks OK.

With these changes there are no remaining occurrences of HOST_TO_NET or NET_TO_HOST in the repository.


jafingerhut (approved):
Looks reasonable to me, with one minor comment.

PNA.mdk
```diff
@@ -376,32 +376,25 @@ here as "main".
...
-| message processing | direction | packets, plus at most |
-|---------------|               | |
-| packet from   | `HOST_TO_NET` | one of: a host-to-net  |
-| host-to-net   |               | recirculated packet,   |
-| recirculate   |               | or one to-net packet.  |
+|               | Processed     | Resulting              |
+| Description   | next by       | packet(s)              |
++:--------------+:--------------+:-----------------------+
+| packet from   | main          | Zero or more mirrored  |
+| network port  |               | packets                |
```
jafingerhut, line 384: In the "Resulting packet(s)" column, would it be more accurate to expand it to something like the following text?

Zero or more mirrored packets, plus at most one copy of the original packet (which may be marked for recirculation).

thomascalvert-xlnx: Are the mirrored packets also allowed to be recirculated? I would think yes, as it may be the only way to support broadcast traffic to a number of VMs larger than the target's maximum mirror slots.

Is the original packet always emitted albeit possibly marked for drop? (The comment in the P4 file says it is dropped after the end of the control block.)

If both the above are true, I would suggest this text:

> Original packet which may be recirculated or marked for drop, and zero or more mirrored packets some of which may be recirculated.

I'm also adding a forward reference to the packet mirroring section (§6.8) in the corresponding paragraph just after this table.

thomascalvert-xlnx:
Text changed as suggested, and forward reference added.

PR #118, opened 2023-04-03
Update command line options to p4test in runp4test.sh to re-enable successful CI

jafingerhut:


jafingerhut:
@mariobaldi Pinging you on this small commit that I merged, so you are aware of it.  CI on this repo tries to use p4test to compile all *.p4 programs in the examples directory, and uses p4test binaries from some docker container that gets re-built with latest version of p4c source code every day or so.  On 2023-Mar-23, a change was made in p4test to enable it to successfully compile P4 programs with arbitrary 'if' statements in P4 action bodies, so the command line options I was using to disable that pass started causing a failure instead.

PR #128, opened 2024-11-23
Switching from Madoko to AsciiDoc for PNA specification

Dscano:
I am currently transitioning the PNA specification from Madoko to AsciiDoc. This is the first PR to initiate the work.


PNA.adoc
```diff
@@ -279,57 +287,61 @@ here as "main".
...
-[^PNAP4TargetSpecific]: It is expected that `pna.p4` include files for
+[1]: It is expected that `pna.p4` include files for
     different targets will be nearly identical to each
     other.  Besides the possibility of differing bit widths for these
     PNA types, the only expected differences between `pna.p4` files
     for different targets would be annotations on externs, etc. that
     the P4 compiler for that target needs to do its job.
 
-For each of these types, the P4 Runtime API[^P4RuntimeAPI] may use bit
+For each of these types, the P4 Runtime API[2] may use bit
```
jafingerhut: I think for a footnote whose content is just a link to some other document, like this one and a few others I will make comments on, it seems preferable to add a bibliography/references section, and use that markup here.

PNA.adoc
```diff
@@ -198,14 +201,13 @@ example). Note that the `main` package is not to be confused with the
 This document contains excerpts of several P4~16~ programs that use
 the `pna.p4` include file and demonstrate features of PNA. Source code
 for the complete programs can be found in the official repository
-containing the PNA specification[^PNAExamplePrograms].
-
-[^PNAExamplePrograms]: <https://github.com/p4lang/pna>
-    in directory `examples`.  Direct link:
-    <https://github.com/p4lang/pna/tree/main/examples>
+containing the PNA specification[1].
```
jafingerhut: This seems better a reference/bibliography entry than a footnote.

PNA.adoc
```diff
@@ -842,9 +861,9 @@ If the property `pna_idle_timeout` is assigned a value of
 `NOTIFY_CONTROL`, the behavior is the same as defined in the Portable
 Switch Architecture if a table has its property `psa_idle_timeout`
 assigned a value of `NOTIFY_CONTROL`.  See the section titled "Table
-entry timeout notification" in the PSA specification[^PSA].
+entry timeout notification" in the PSA specification[1].
```
jafingerhut: Please replace with a reference/bibliography entry.

jafingerhut:
Please ping me when you believe this is ready to go.  It looks like there are still work on tables to be done, and probably some other things.

Dscano:
I propose using [NOTE ](https://docs.asciidoctor.org/asciidoc/latest/blocks/admonitions/#admonition-types)  to replace Begin/End TBD

jafingerhut:
> I propose using [NOTE ](https://docs.asciidoctor.org/asciidoc/latest/blocks/admonitions/#admonition-types) to replace Begin/End TBD

Sounds reasonable to me.  Go for it, and we can take a look at the PDF results, but I'd be surprised if there is any problem with the idea.

Dscano:
@jafingerhut I have finished converting  PNA specification from Madoko to AsciiDoc.

- Figures 2 and 3 are not present in the `PNA.mdk`, so I don't know how to map them in `PNA.adoc`.
- In my opinion, it's worth developing the HTM version as well.
- It's worth developing the same automation mechanisms used in p4runtime and p4-spec.

jafingerhut:
There are two sections in the table of contents named "Revision history".  Ideally there should be only one, which should be the first appendix.

jafingerhut:
The current Madoko PDF has appendices A, B, C, etc.  As of the time of writing this comment, the AsciiDoc PDF has one appendix, A, with separate sub-sections for each of what were appendices A, B, C, ...

If it is straightforward to make the AsciiDoc appendices be more like the original Madoko ones, that would be nice.  I believe it is possible, as I think we did this for the P4 language spec AsciiDoc.

PNA.adoc
```diff
@@ -168,38 +51,39 @@ The Portable NIC Architecture (PNA) is the P4 architecture that defines
...
 
-2. A library of types (e.g., intrinsic and standard metadata) and
+* A library of types (e.g., intrinsic and standard metadata) and
 P4~16~ externs (e.g., counters, meters, and registers).
 
 PNA is designed to model the common features of a broad class of NIC
 devices. By providing standard APIs and coding guidelines, the hope is
 to enable developers to write programs that are portable across
-multiple NIC devices that are conformant to the PNA[^PNAPortability].
+multiple NIC devices that are conformant to the PNA[1].
```
jafingerhut: Insert whitespace before `[1]`.

jafingerhut: We might also want to use different text than `[1]`, since there are also auto-generated bibliography references that look like `[1]`, too.  Maybe `[Footnote 1]`, both where it is referenced from, and where the footnote is shown?

PNA.adoc
```diff
@@ -356,101 +238,97 @@ In this document we use the following naming conventions:
...
+[#sec-pna-type-definitions]
+=== PNA type definitions 
 
 Each PNA implementation will have specific bit widths in the data
 plane for the types shown in the code excerpt of Section
-[#sec-pna-type-definitions-code].  These widths are defined in the
+<<#sec-pna-type-definitions-code>>.  These widths are defined in the
 target specific `pna.p4` include file.  They are expected to differ
-from one PNA implementation to another[^PNAP4TargetSpecific].
+from one PNA implementation to another[1].
```
jafingerhut: I would recommend `[Footnote 2]` here, to avoid duplication with the earlier `[Footnote 1]`.

jafingerhut:
Thanks for updating the section headings in the appendices.

I would recommend removing the word "Appendix" from the AsciiDoc source for the titles of those appendices, because as of commit 5, the generated PDF output has titles like these for the appendices:

`Appendix A: Appendix: Revision History`

I believe the `Appendix A` is auto-generated, and looks fine to me, but the `Appendix:` part is from the AsciiDoc source, because Madoko did not auto-generate the word "Appendix" in the title.

Makefile
```diff
@@ -1,15 +1,17 @@
 SPEC=PNA
+ROUGE_STYLE=github
 
-all: build/${SPEC}.pdf
+all: ${SPEC}.pdf
 
-build/${SPEC}.pdf: ${SPEC}.mdk
-	madoko --pdf -vv --png --odir=build $<
-
-build/${SPEC}.pdf: p4.json
```
jafingerhut, line 8: I believe the p4.json file should be removed with this PR, unless you find some reference to that file in the files that remain.  I think p4.json was only used for Madoko syntax highlighting.

Makefile
```diff
@@ -1,15 +1,17 @@
...
-build/${SPEC}.pdf: pna.p4
+${SPEC}.pdf: ${SPEC}.adoc
```
jafingerhut: Please add the file pna.p4 as a dependency on this rule, and the rule for generating HTML output, since PNA.adoc has include directives that include portions of pna.p4.  If pna.p4 changes, it is likely that the PDF and/or HTML files need to be updated, too.

jafingerhut: Also, please add a Makefile rule to generate HTML output.  I don't see one here.  It would be helpful to have that, even if the original Makefile did not have such a rule to generate HTML from Madoko source.

jafingerhut (approved):
LGTM

jafingerhut:
I will wait to merge this after mentioning this change at the next arch work group meeting on 2024-Jan-13.

PR #131, opened 2026-04-23
Add REUSE-lintable copyright and license info to all files

jafingerhut:


jafingerhut:
All files are documentation and PNA specification related.  All are licenced as Apache-2.0.  No uses of scapy or any other copylefted packages in a way that requires files here to be licensed with a copyleft license.

